PR #2, opened 2015-10-03
Document hash locationType config requirement

sohara:
I thought it would be useful to let people know that they need to use the 'hash' location type. Also a way of supporting both Electron and regular ember-cli builds with History API support.


felixrieseberg:
Thanks for the PR, I can totally see how people might be confused without it. Thanks!

PR #5, opened 2015-11-28
Use loadURL instead of (deprecated) loadUrl

romulomachado:


felixrieseberg:
Oh nice, thank you very much!

PR #11, opened 2015-12-26
Enables Node.js mode

twokul:
Silences jshint warnings about node variables (`require`, `process`)


felixrieseberg:
Good catch, thank you very much :clap:

PR #13, opened 2015-12-29
Updates to the latest electron

twokul:


felixrieseberg:
Thanks a lot, good catch!

PR #27, opened 2016-02-12
Option to set ember build environment when packaging

nsantorello:
Allows `ember electron:package --environment=X` to build the package with `environment` set to `X`.  Currently it is hardcoded to `production`:
https://github.com/felixrieseberg/ember-electron/blob/master/lib/commands/package.js#L141

### DIFF
--- a/lib/commands/package.js
+++ b/lib/commands/package.js
@@ -20,7 +20,7 @@ module.exports = {
     availableOptions: [{
         name: 'environment',
         type: String,
-        default: 'development',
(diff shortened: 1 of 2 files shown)


felixrieseberg:
Oh nice, even includes tests - thank you very much, awesome contribution :clap:! I'll push out a new release as soon as the tests pass.

felixrieseberg:
Actually, tests are borked anyway. I'll merge and fix tests.

PR #34, opened 2016-02-15
Add Proper Credit

brzpegasus:
Given that a significant percentage of this addon's code was copied directly from [`ember-cli-nwjs`](https://github.com/brzpegasus/ember-cli-nwjs), namely:

* serve & test commands
* test task/runner/custom TAP reporter
* helpers
* shims
* all associated unit tests

with the exception of fixing the indentation and replacing `nw` with `electron`, I thought it wouldn't be unreasonable to ask for credit in the license, similar to what you did for `ember-cli-remote-inspector`.


felixrieseberg:
Totally reasonable. I never meant to hide that (it's in the readme), but this makes a lot of sense. Thank you very much!

felixrieseberg:
I'll push out a new version too, so that peeps get the new license file 🌈

PR #35, opened 2016-02-16
only inject socket.io when EMBER_CLI_ELECTRON = true

curtisgr:
addresses "Uncaught ReferenceError: io is not defined" when running the standard development environment using `ember server` where ember-electron package is also present in project.

### DIFF
--- a/index.js
+++ b/index.js
@@ -110,14 +110,14 @@ module.exports = {
             return injectScript('shim-footer.js');
         }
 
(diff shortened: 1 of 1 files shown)


felixrieseberg:
Cool, thanks!

nsantorello:
:+1: came here for this, @curtisgr thanks!

felixrieseberg:
Thanks again - this is now in 0.5.7 :+1:

curtisgr:
No problem, thanks for the quick turn around!

Issue #4, opened 2015-11-05
Development server?


sohara:
A few questions. On this line: https://github.com/felixrieseberg/ember-electron/blob/master/blueprints/ember-electron/files/electron.js#L21, there is reference to a development server on port 5000. This doesn't seem to run by default?

Also, most ember apps involve API requests to server, often with a relative path. That appears to breaks since with ember-electron we are dealing with a `file://` protocol? Is there a way to get around that, that you know of?

felixrieseberg:
Hey, the dev server is only invoked if you set the environment variable right in front of it - otherwise, it'll go with the file protocol. If you want to get around that, you could try working with `base`?

PR #45, opened 2016-03-22
Enable express server on development

calderas:
I was playing around integrating the server on an app and this did the trick for development.

Wondering if this might be useful to add.

related to #4

### DIFF
--- a/lib/commands/electron.js
+++ b/lib/commands/electron.js
@@ -4,6 +4,8 @@ const spawn = require('child_process').spawn;
 const chalk = require('chalk');
 const RSVP = require('rsvp');
 const debugServer = require('../helpers/debug-server');
+const existsSync = require('exists-sync');
+const path = require('path');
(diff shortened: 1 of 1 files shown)


SSS2557:
this is wonderful :)

Does nobody else feel this is required?

Or is this just waiting for the tests to pass?

felixrieseberg:
Hey, the reason I haven't merged this is that I'm super conflicted about adding this for everyone who's using Ember-Electron. I'm thankful for any PR, but adding a server to your app is an anti-pattern the whole Electron team is pretty strongly against - it will lead to insecure apps with unnecessary overhead.

If someone needs a server during development, it should probably really just be a separate process.

SSS2557:
oh :) @felixrieseberg 

Thanks for letting know. That makes sense. But, would you happen to have any insights as to what is the best way that can be done?

Would also be very grateful if you can provide more direction to it :) 

Right now, I am only able to get our app up by pointing main to electron.js

But we do have a couple of needs to go to express-server pre-load

PR #48, opened 2016-03-24
Package command: ensure target exists

AlexCharlton:
Packaging on OS X failed silently for me without ensuring that the target exists.

I hope this is the right solution to this problem!


felixrieseberg:
Good call, thanks for the PR!

PR #53, opened 2016-04-16
added optional dev dependency copying

devlo:
We check in installDependencies if node module from dependencies is inside devDependencies, after that we push it to array ( before in this point i would call copyDevModule to copy one module but because of recompiling native dependencies it was changed). 

After recompileDependencies we call copyDevModules to copy those modules.


felixrieseberg:
Thank you very much, great stuff! I'm sure a bunch of users will appreciate your two features 👍

felixrieseberg:
...and published as 1.4.0 to npm!

devlo:
Nice! Hope it helps other users too. My first real contribution on Github :)

felixrieseberg:
Congrats, it's a good one! Be careful - this whole open source thing is super addictive, but it's fun 😜 Keep doing it, the world depends on contributors like you!

PR #85, opened 2016-08-01
doc/config Add commented block to allow hash and auto types in config…

dirvine:
…/environment.js.

This is not important and only a convienience

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -1,10 +1,16 @@
 /* jshint node: true */
 
-module.exports = function(environment) {
+module.exports = function (environment) {
   var ENV = {
     modulePrefix: '<%= dasherizedPackageName %>',
     environment: environment,
     baseURL: '/',
+    // Uncomment below to allow "ember serve" & "ember electron" to both run.
+    // Note ember serve must be run first at this time.
+    // Security must be considered in this case, but may prove useful
+    // in testing and prototyping.
+    //
+    // locationType: process.env.EMBER_CLI_ELECTRON ? 'hash' : 'auto',
     locationType: 'hash',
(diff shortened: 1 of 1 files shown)


felixrieseberg:
Looks really nice! One quick question though: I went with Ember's style and used semicolons - could you add those back?

If not, no worries, I can also merge and add them back myself right after.

dirvine:
Ah no prob, just give me 2 mins.

dirvine:
Actually @felixrieseberg  I am not clear what you mean here. Sorry, I was thinking about line endings, but this is not. Perhaps to save time if you want to merge and fix I can see what you mean. otherwise I am happy to change here and push a commit over this if you give me a bit more info.

felixrieseberg:
No worries, I can take care of it. And yes, I just meant the `;` at the end of lines :wink: I don't really care about one code style over the other, but since [ember-cli uses them in their config](https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/config/environment.js), I think it'd be fair to keep them around.

dirvine:
Ah thanks for that and sorry for not spotting it. vscode eslint or similar, I must change that setting. 

BTW this is a brilliant project and for newcomers to ember it's really compelling to see the small gap (sans security worries as you spoke about etc.) between the rendering of desktop and web apps. I have never worked in front end so it's pretty new all the MV* stuff and this all helps to see something fast and get a feeling of understanding the differences of desktop/web apps. 

Anyway I just wanted to say thanks. Really neat.

felixrieseberg:
Thanks for your kind words, much appreciated!

PR #102, opened 2016-10-11
Fix app declaration

lucascoupez:
Fixes the following error in electron.js:
const app, BrowserWindow   = electron.app;
      ^^^
SyntaxError: Missing initializer in const declaration


felixrieseberg:
Thank you!

PR #103, opened 2016-10-11
:bug: Fix lookup of prebuilt package

drewchandler:
The package command fails when it recompiles native dependencies.

There are 2 issues:
* `getElectronPackagePath` expects a project object not a path
* it looks for `package.json` in `electron-prebuilt`'s `dist` directory rather than its root

I've fixed these issues and packaging works for me again.

### DIFF
--- a/lib/commands/package.js
+++ b/lib/commands/package.js
@@ -315,8 +315,8 @@ module.exports = {
       }
 
       const rebuild = require('electron-rebuild')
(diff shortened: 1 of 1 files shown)


drewchandler:
Looks like Travis had some connection issues 😭 I'll force push a new commit to trigger a rebuild.

felixrieseberg:
Thank you very much!

PR #120, opened 2017-01-12
[WIP] Update testem integration

bendemboski:
## Overview

ember-electron does not play nicely with newer versions of ember-cli and testem. #118 is one example of a bug that exists, but also the code that gets injected into the browser window no longer fully speaks the same language as testem, so even after fixing #118, running in dev mode (`ember electron:test -s`) is kinda funky (it spawns a second testem tab and the results appear there, with the first one remaining blank). Also, ci mode doesn't produce the full set of output (logs and whatnot) that `ember test` produces.

## ~~Open~~ Resolved questions

1. ~~~What kind of backwards compatibility do we need? I can't tell from `ember-try.js` or `package.json` what version of ember-cli this is actually meant to support. I need to investigate, but I suspect with fairly modest effort we can support all released 2.x version (>= 2.3.0 -- earlier 2.x releases never made it out of beta). It will probably take some more significant effort to support 1.13.x.~~~ This supports all released 2.x versions, but not 1.13.
1. ~~~Since we depend on some implementation details of testem, should we pin a version of it in our package.json's dependencies? Would that even work?~~~ We pin a version of testem and make sure the use that pinned version when running `electron:test`.
1. ~~~Can we make a small change to testem to remove the need for our knowledge if its implementation details? I think so, but need to investigate.~~~ My PRs were accepted into testem, so we are no longer assuming any implementation details.

## Benefits of this change

* After a bit of URL re-writing, all communication with testem happens in the same way as it happens when running `ember test` using testem's code, so no custom output parsing/TAP generation, no custom QUnit listening code, and all of the same output is produced as `ember test`.
* The `electron:test` command is now a sub-class of the `ember test` command, and contains very little custom code that needs to be maintained
* The testem configuration is moved to a config file (`/testem-electron.js`) so it can be configured in the same way as `ember test`
* The testem implementation is primed for any of the integrations testem offers (jasmine, mocha, buster)
* When testing, all of the `copy-files` files are included, so modules can be shared between `electron.js` and `tests/electron.js` so tests can run against some form of the main process.

## My next steps

- [x] Test on other ember-cli versions to determine when compatibility breaks
- [x] Gather feedback on which versions we should keep supporting
- [x] Implement support for those version
- [x] Investigate a change that I have in mind that, if merged into testem, would allow us to remove pretty much all of our knowledge of testem's internal implementation
- [x] Write & update unit tests
- [x] Update documentation

PR #129, opened 2017-01-30
[WIP] Use electron-forge under the hood

anulman:
## TODO

- [x] update blueprint to use `forge#import`
- [x] update electron cmd to use `forge#start`
- [x] update package cmd to use `forge#package`
- [x] rename and update ship cmd to use `forge#make`
- [x] review & rm `// todo`s


malept:
🎉 So excited for this!

When I was doing a prototype for a similar kind of PR (that I am officially abandoning in favor of this :smile:), I noticed that there's an undocumented `ship` command that appears to be the same as/similar to `electron-forge make`.

anulman:
@malept Good catch! That was my old shim ;)

Added to the todos. I'm most excited for publish though 🙏

package.json
```diff
@@ -65,6 +65,7 @@
     "broccoli-merge-trees": "^1.1.0",
     "broccoli-string-replace": "^0.1.1",
     "chalk": "^1.1.0",
+    "electron-forge": "isleofcode/electron-forge",
```
anulman: TODO: revert to mainline electron-forge once https://github.com/electron-userland/electron-forge/pull/113 merges

malept: Electron Forge 2.8.0 is ready for your consumption 😄

anulman: 🎉 on it once behind a terminal

anulman:
Lol, we disagree on choice of linter / config 😛

anulman:
Should be good for merge + npm beta release.

Remaining items I'm tracking before release (@felixrieseberg want to track in a meta-issue?):
- [ ] Support `--arch`, `--platform`, `--targets`, `--skip-build`, `--skip-package` options in `package` & `make` commands
- [ ] Ensure `ember electron:test` still works
    - n.b. I can't do this on my base project, as IIRC `ember electron:test` does not work with ember-mocha
- [ ] Update readme w/ new package.json config guides
- [ ] Update test suite

And some lib improvements I'd noted in todos (each could / should be its own issue, if adopting):
- [ ] Consider replacing ember config rewrite (i.e. `rootUrl` & `locationType` handling) w/ a validator that fails if not set to hash
- [ ] Consider if InspectorServer can work w/ configurable port values
- [ ] Tidy up InspectorServer

package.json
```diff
@@ -65,6 +70,7 @@
     "broccoli-merge-trees": "^1.1.0",
     "broccoli-string-replace": "^0.1.1",
     "chalk": "^1.1.0",
+    "electron-forge": "^2.8.0",
     "electron-packager": "^8.3.0",
```
malept: This is now an implicit dependency, unless for some odd reason we need this as a direct dependency.

.travis.yml
```diff
@@ -4,8 +4,8 @@ language: node_js
 sudo: false
 
 node_js:
-  - "4.2"
```
malept: The readme should now state that this package requires Node 6.

package.json
anulman: Good catch. There are several others that can be cleaned out too, like winstaller / ect

.travis.yml
anulman: Doesn't it? https://github.com/felixrieseberg/ember-electron/blame/feat/forge/README.md#L16

malept: Ah, the Travis config confused me.

kevingelion:
Will this support signing still? I see `electron-forge` has a `publish` command and curious if that will make its way in here.

malept:
Signing (currently only macOS) is part of `electron-forge package`, as it's functionality provided by Electron Packager (via electron-osx-sign).

anulman:
@kevingelion check this PR; we're extending support in forge for not-current-platform installers: electron-userland/electron-forge#144

I've not tested with signing creds yet, though they should be supported via `electronPackagerConfig` within the forge config: electron-userland/electron-forge#144

kevingelion:
@anulman i'm guessing this would eventually make its way into `ember-electron` then? 😄

anulman:
@kevingelion it's already working in this branch; you just need to monkeypatch which forge version you're consuming ;)

Happy to help with this later if you'd like to test!

felixrieseberg:
:tada: Let's merge it - all additional work can happen in additional PRs 👍 

@anulman You're a star 🌟

lib/commands/build.js
```diff
@@ -0,0 +1,206 @@
+'use strict'
+
+const RSVP = require('rsvp')
+
+const fs = require('fs-extra')
+const glob = require('globby')
+const npmi = require('npmi')
+const path = require('path')
+
+const Command = require('ember-electron/lib/commands/-command')
```
bendemboski: @anulman @felixrieseberg am I missing something here? Using this kind of absolute require breaks for me locally, and is also causing the travis tests to fail. Why aren't all of these relative paths?

anulman: Nope, dumb error. Feel free to fix, or I'll hit it in my next sprint :)

bendemboski: Okay, added to #120

lib/commands/electron.js
```diff
@@ -1,127 +1,121 @@
...
+      .then(() => this._runInspectorServer(options, logger))
+      .then(() => {
+        logger.message('Starting Electron...')
+        logger.line()
 
-  runInspectorServer: function (options) {
-    if (options.environment !== 'development') {
-      return
-    }
+        return start()
```
bendemboski: @anulman Something is baffling me here as I'm trying to figure out how to get `ember electron:test` running through electron forge, rather than requiring electron to be installed in the host app. This `start()` call doesn't have any options, and I don't see any chdirs, so how does electron-forge know to run against the output of the build command?

bendemboski: Ah, never mind, figured it out.

Issue #133, opened 2017-02-07
ember electron:build removes cursor in terminal


halfbyte:
This is somewhat of an annoyance: Whenever I run said command, my cursor is gone and can only be restored by a `reset` or by doing an `echo "\u001b[?25h"`.

I've tried to look at the build command and I can fix the bug by inserting `this.ui.stopProgress()` into the [cleanup](https://github.com/felixrieseberg/ember-electron/blob/8c1f936b5c8cd09a8612083926c70a8b68c67e84/lib/commands/package.js#L361) function.

I'm not sure this is the correct fix, though. I tried to remove the `startProgress` call from the [build()](https://github.com/felixrieseberg/ember-electron/blob/8c1f936b5c8cd09a8612083926c70a8b68c67e84/lib/commands/package.js#L175) function, but that doesn't fix it, so maybe there are other occasions where something like `startProgress` is called. 

In any case, looking at the code of `startProgress` and the underlying ora code, calling startProgress without a corresponding `stopProgress` is probably not a good idea.

felixrieseberg:
Hey, thanks for the issue and the research! Here's my take: I agree that we should have the `stopProgress` and if it fixes an annoyance for you, it deserves to be in. Want to file a quick PR? I'd hit ✅ on it right away :tada:

PR #134, opened 2017-02-08
call stopProgress() in cleanup to reenable cursor after packaging

halfbyte:
This PR delivers a fix for #133.

The fix is working, but I still think that someone else in the used packages does something bad as well, as removing the startProgress() did not fix the issue. Will try to investigate further.

I've tried to add a little test, I hope the way I did it is roughly compatible with the way you write tests over here.

Thanks for the encouragement, @felixrieseberg, that's how you get PRs :)

### DIFF
--- a/lib/commands/package.js
+++ b/lib/commands/package.js
@@ -358,6 +358,7 @@ module.exports = Command.extend({
 
   cleanup: function () {
     return new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


felixrieseberg:
Thank you so much, merged right away!

felixrieseberg:
Released as `1.12.6`!

Issue #143, opened 2017-02-23
Adopt Ember project codestyle


pichfl:
Ember is switching away from JSHint to ESLint and now provides plugins for a more consistent code style.

To my understanding this project deviates from typical ember addons and uses StandardJS, which is as of now not quite prepared to handle modern JS syntax used by Ember (let/const, arrow functions, etc are still topics to be discussed by its maintainers).

My suggestion is to follow the code style of the ember project. Since this is mostly cosmetic, I would be happy to provide a PR if you like the idea, or just shut up if you don't. 😁

anulman:
+1. @felixrieseberg?

felixrieseberg:
Into it! Let's merge the current electron-forge work so we got a clean slate to start the work. @pichfl, a PR would be extremely welcome <3!

pichfl:
Will do 🙂

bendemboski:
#120 will cause you some merge headache also if you do this off of master. I think it's ready to merge, but I'm hesitating because of failing unit tests in master... @felixrieseberg should I just go ahead and merge? I *think* it all works...

pichfl:
I'll wait for #120 then.

PR #144, opened 2017-02-24
WIP: Move from JSHint/StandardJS to ESLint with sensible configuration

pichfl:
* code style based on ember-suave eslint plugin
* should provide a more consistent code style in the Ember Addon landscape
* conversion was done mostly automatically using `eslint --fix`, artefacts may occur

- [ ] Fix tests
- [ ] Discuss

Re #143


.eslintrc.js
```diff
@@ -0,0 +1,49 @@
...
+    // 'Symbol': true,
+    // 'WeakMap': true,
+  },
+  rules: {
+    'comma-dangle': ['error', 'always-multiline'],
+    'ember-suave/no-const-outside-module-scope': 'off',
+    'ember-suave/no-direct-property-access': 'off',
+    'ember-suave/require-access-in-comments': 'off',
+    'newline-before-return': 'error',
+    'new-cap': 'off',
```
bendemboski: Why turn this rule off?

pichfl: Had to start somewhere. Turning those on throws quite a few errors in the project.
We can gradually switch them on when the transition is done.

For explantation
- [ember-suave/no-direct-property-access](https://github.com/DockYard/eslint-plugin-ember-suave/blob/master/docs/rules/no-direct-property-access.md) needs a lot of manual adjustments. Should however not be that hard to enable in the long run.
- [ember-suave/require-access-in-comments](https://github.com/DockYard/eslint-plugin-ember-suave/blob/master/docs/rules/require-access-in-comments.md) enforces JSDoc documentation guidelines unused in this project at the time
- [new-cap](http://eslint.org/docs/rules/new-cap) needs very fine tuning as Electron exposes quite a few objects with capital letters which are not used as constructors. Can be done, just a bit much for a quick test.

bendemboski: Yeah, totally -- definitely with code style updates do fast & iterative changes to minimize merge conflicts :)

All of the others seem fine to me, and I think a number of Ember folks are backing off of the Ember destructuring practice (`no-direct-property-access`), although I still like it a lot. Agree that `require-access-in-comments` doesn't make sense for a project that doesn't expose a Javascript API. I have no strong feelings about `newline-before-return`.

`new-cap` is the one I was questioning -- maybe electron has a bunch of capital letter functions (I don't think I've personally run into them), but I haven't seen much code in *this* project that would violate the `new-cap` rule. Am I missing something -- does it actually cause more than just one or two errors?

pichfl: You're right on `new-cap`, it actually triggers only once. 

In `lib/helpers/debug-server.js` L14
```js
const remoteDebugServer = require('http').Server(remoteDebugger);
```
(well, thanks Node)

Will update with a inline disable for just that line right now.

bendemboski: Sweet :) Otherwise, looks great to me! Nice work! And now I won't have to create a new code style configuration in my IDE to be able to work on ember-electron without hating life!

pichfl: Pretty much the reason why I started this. Our own guidelines are far more strict than this (based on AirBnB), but do not fit the Ember community, so I think this is a solid compromise while stile providing some consistency. 

There is stuff like [prefer-arrow-callback](http://eslint.org/docs/rules/prefer-arrow-callback) which I would gladly see here in all electron files, but won't work for Ember code, so I'm not sure if we should include it here. Still, having a very modern node project gives us a lot of nice syntax.

Issue #141, opened 2017-02-22
shim-head and shim-footer should only be injected if running in Electron


pichfl:
As of today, ember-electron pollutes the index.html even for regular builds of the application using `ember build`. 

**Expected behaviour**
The require / requireNode shims should only be included if the application is running inside of Electron.

Basically wrapping the whole contentFor() method in index.js in a suiting conditional should do the trick.

anulman:
👍 I like. Okay if we wait until #129 is merged?

Also, think you could write this PR?

pichfl:
Yes and yes.

PR #145, opened 2017-02-24
Inject shim-head and shim-footer only in Electron environment

pichfl:
* No longer pollutes index.html if the project is compiled outside of ember-electron
* Use fancy ES6 destructuring

Re #141

### DIFF
--- a/index.js
+++ b/index.js
@@ -82,11 +82,13 @@ module.exports = {
   },
 
   contentFor(type) {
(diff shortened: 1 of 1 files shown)


pichfl:
Could be merged before #144 but should wait for the tests to be fixed. I manually tested on two of my projects but it might be worth to add an acceptance test for this?

PR #146, opened 2017-02-25
Fix/improve ember electron:test

bendemboski:
* Fix a logical merge conflict between #120 and #129 that was causing us to not copy the right files during a test build
* Switch over to using electron-forge for running electron:test
* Add a node library included from the main process so the ember electron:test step will test the functionality that copies it to make it available in tests.


bendemboski:
@felixrieseberg or @anulman one of you want to take a quick look? Turns out I didn't quite do all the needed updates before my testem PR was merged, so this is some quick cleanup.

pichfl:
Sorry for messing up your PR. I'll take a look and fix what I broke.

bendemboski:
@pichfl thanks for the offer, but I've just about got the conflicts cleaned up already, so I'll take care of updating this PR 😄

bendemboski:
@pichfl @felixrieseberg PR is updated, and it was very satisfying to be using the newer Javascript style 😃

PR #149, opened 2017-02-27
Prevent infinite loop within Command init()

kevingelion:
While debugging a recent install of 2.beta, @anulman and I came across a strange loop within `-command.js`:

### DIFF
--- a/lib/commands/-command.js
+++ b/lib/commands/-command.js
@@ -6,7 +6,7 @@ module.exports = Command.extend({
   init() {
     process.env.EMBER_CLI_ELECTRON = true;
 
-    if (this._super && this._super.init) {
+    if (this._super && this._super.init && this._super.init !== this.init) {
       this._super.init.apply(this, arguments);
     }
   },
--- a/lib/commands/-command.js
+++ b/lib/commands/-command.js
@@ -5,9 +5,6 @@ const Command = require('ember-cli/lib/models/command');
 module.exports = Command.extend({
   init() {
     process.env.EMBER_CLI_ELECTRON = true;
-
-    if (this._super && this._super.init && this._super.init !== this.init) {
-      this._super.init.apply(this, arguments);
-    }
+    this._super(...arguments);
   },
 });


lib/commands/-command.js
```diff
@@ -6,7 +6,7 @@ module.exports = Command.extend({
   init() {
     process.env.EMBER_CLI_ELECTRON = true;
 
-    if (this._super && this._super.init) {
+    if (this._super && this._super.init && this._super.init !== this.init) {
```
bendemboski: Can't this whole block just be `this._super(...arguments)` (or `this._super.apply(this, arguments)` if we care about older Node compatibility)?

bendemboski: That's what I've been doing [here](https://github.com/felixrieseberg/ember-electron/blob/master/lib/commands/electron-test.js#L12), and it's working fine AFAIKT.

anulman: That would also be my preference; I vaguely remember trying that at one point but there was a lot of churn so ¯\_(ツ)_/¯

kevingelion: Ideally, I would also prefer to use your suggestions but before making this PR i tried using ES6 syntax (identical to what you're doing @bendemboski) and was met with:
```Function.prototype.apply was called on #<Class>, which is a object and not a function
TypeError: Function.prototype.apply was called on #<Class>, which is a object and not a function```

kevingelion:
Worthwhile to note that I also could reproduce this on Windows x64.

lib/commands/-command.js
bendemboski: @kevingelion are you sure you were running against the correct ember-cli version? When I change to that syntax, everything works fine for me locally. Do you have a more specific repro?

If this is a problem here, then it's probably a problem in `electron-test.js` also, so we probably want to fix it in both places, but I'd prefer to understand why this pretty fundamental mechanism of ember-cli's `CoreObject` is misbehaving for us...

anulman: Total conjecture: could my issues have stemmed from using an older e-cli version; and @kevingelion's from keeping the `.apply(...)` syntax?

bendemboski: Plausible -- I don't remember exactly how, but I'm pretty sure in older version of ember-cli invoking super-class functions was more fiddly.

kevingelion: Ah, okay. I'm running ember-cli `2.4.3`. My fault for not even considering that...

anulman:
@kevingelion I pushed the change discussed inline & am about to merge. If the behaviour recurs, we know what to do :)

PR #156, opened 2017-03-03
Use a ember protocol for file loading

pichfl:
- No longer requires changing the locationType
- caches file system lookups (simple but should do the job)
- [x] Check if this works with locationType:hash

I split registering the Scheme and the protocol in two intentionally to make it easier for developers to register additional protocols.


anulman:
Yeah! I like your solution much better than the validations

pichfl:
I recommend to merge #155 and #165 before this, but it should work as expected. My understanding for the testing environment is not good enough to include the same thing in the tests in `node-tests/fixtures/project-with-test-config/tests/electron.js` but if someone points me in the right direction I'll file a separate PR to fix those as well.

kevingelion:
This is great! I was having issues rendering a THREE.js scene and couldn't figure out why last week and felt like it was related to relative asset paths or something and turns out this branch works like a dream. Can package/make successfully on ember 2.8!

pichfl:
I'm still in discussion with @bendemboski about details regarding testing, but I'm happy to hear it works as expected.

lib/commands/build.js
```diff
@@ -147,6 +147,7 @@ module.exports = Command.extend({
     let ee = this.project.pkg.config['ember-electron'] || {};
     let patterns = ee['copy-files'] || [
       'ember-electron/electron.js',
+      'ember-electron/protocol-ember.js',
```
bendemboski: No longer needed, right?

package.json
```diff
@@ -110,6 +111,7 @@
     "ember-electron": {
       "copy-files": [
         "ember-electron/electron.js",
+        "ember-electron/protocol-ember.js",
```
bendemboski: Also no longer needed?

lib/commands/build.js
pichfl: Aye.

package.json
pichfl: 1 AM artifact.

felixrieseberg:
:rocket:

PR #165, opened 2017-03-07
Updated shims

pichfl:
The whole shims topic kept on nagging me, so here is a small, mostly cosmetic update.

* Modern JS syntax (supported since Chrome 45/46/47)
* Added a few comments to clear up where the injections come from and what they actually do

Changes to index.js:
* reading file in script injections separated in variable for readability
* collapsed conditional for injections (this was an artifact from earlier versions)

Changes to head:
* Use modern code style (arrow functions)
* Always export win.ELECTRON as the file is only injected in electron
* added comments
* window global is only called once

Changes to footer:
* Use modern code style (arrow functions, destructuring, rest params, spread)
* Cleaned up try/catch/if/else paths (no need for else if we return early)
* added comments

### DIFF
--- a/blueprints/ember-electron/files/ember-electron/electron.js
+++ b/blueprints/ember-electron/files/ember-electron/electron.js
@@ -32,8 +32,6 @@ app.on('ready', function onReady() {
         height: 600
     });
 
-    delete mainWindow.module;
-
     // If you want to open up dev tools programmatically, call
     // mainWindow.openDevTools();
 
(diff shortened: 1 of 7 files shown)


pichfl:
I'm wondering @anulman @felixrieseberg do we actually need the `delete mainWindow.module;` line in our blueprints any longer? Shouldn't this be fine by the way our shims work?

felixrieseberg:
I think there are still a few modules out there with older loaders that can get confused if you don't have it in. We could easily take it out, add a `// Todo` in code and remind ourselves that we took it out once people complain.

This looks otherwise ✨, I'm merging it!

Issue #123, opened 2017-01-18
Diferent copy-files by platform


debelop13:
Hello!

Can be possible to add different files (depending on platform) in packaging?

Thanks.

danidr:
+1. I would also love to know this, since I am including a quite big (~500MB per architecture) external executable and including all platforms in all packages would make for a huge application.

debelop13:
Hi! 
Finally I achieved this using --ignore like in this case:

ember electron:package --platform=darwin --ignore=folderToIgnore

ember electron:package --platform=win32 --ignore=folderToIgnore

felixrieseberg:
Seems like a solid idea. If `copy-files` was also a command line parameter, your use case would be covered, right?

Would you be interested in making a PR?

danidr:
Well, seeing that this issue will come up quite often, wouldn't it be better to create a sort of way to filter the `copy-files` options?
This could be extended to any type of setting, for example by creating a sub-category in the package.json for every platform (which could of course be optional, the property id could be set to 'all' and work for any type of platform). Like:
`
'win32': { options... },
'darwin': { options... }
`
At this point though, it would make more sense to create a config file specifically for ember-electron.

anulman:
Hm. What if we relied on the fs directly?

#129 adds the notion of the `ember-electron` dir (since electron.js can't be at the same level as package.json). What if we concat'ed trees, broccoli style?

- `ember-electron/public/:platform` tree, merged into
- `ember-electron/public`; tree, merged into
-  `dist/public`

bendemboski:
@anulman I like the idea of using the fs, but I think the platform sub-directory of `public` might be confusing. What do you think about making the platform public directories siblings of the "base" public directory:

- `ember-electron/public-:platform` tree, merged into
- `ember-electron/public`; tree, merged into
- `dist/public`

anulman:
@bendemboski I've really liked your dash idea, but on second thought want to propose one more amendment: what about `-:platform` subdirs instead?

I think this better maps to a lot of Ember fs styleguides I've seen, and has the added benefits of being an easily globbable rule (`!.../-*/**/*`) + puts these special dirs up at the top of alphabetical lists.

To recap, the structure I'm proposing is:

- `ember-electron/public/-:platform/**/*`; merges into
- `ember-electron/public/**/*`, merges into
- `$packagedAppFiles/dist/public`

bendemboski:
[...]

PR #169, opened 2017-03-08
Use `public` folders instead of `copy-files`

jacobq:
Here's an attempt at implementing the changes proposed by #123 
Note that tests & docs still need to be updated.


jacobq:
Closing in favor of https://github.com/felixrieseberg/ember-electron/pull/170

PR #170, opened 2017-03-09
Replace copy-files with public directories

bendemboski:
Instead of listing files in a config block in package.json, we now pull them from the filesystem. Platform-independent files go in ember-electron/public/, and platform-specific files go in ember-electron/public-<platform> (and overwrite platform-independent files at the same relative path).

This is implemented as part of the broccoli pipeline now rather than in our custom cli commands so that it doesn't have to be re-implemented in each of the few different ways we build the app for our various commands.

This change necessitated a more explicit packaging of the electron application when running ember electron. That command used to just run the electron app out of the project in place (i.e. launch /ember-electron/electron.js). But this is no longer possible with the public directory merging. So, the broccoli pipeline puts the ember app in dist/ember, and the electron app in dist/electron, leaving dist/ nice and clean so we can put package.json there and have it actually look like the nice clean project that ember forge expects. This same directrory structure is created in a temp directory when running ember electron:test, ember electron:package and ember electron:make.

Still TODO:
- [x] Update README to describe new filesystem
- [ ] Move forge config out of package.json per @anulman's suggestion
- [x] Rebase onto #156 and resolve code/logical conflicts


index.js
```diff
@@ -82,6 +128,30 @@ module.exports = {
...
+  postBuild({ directory }) {
+    if (!process.env.EMBER_CLI_ELECTRON) {
+      return tree;
+    }
+
+    //
+    // electron-forge needs our package.json to have a main entry pointing to
+    // electron.js. I can't figure out how to get broccoli to edit JSON for us,
+    // so we just do it manually once the build is complete.
+    //
```
jacobq: @bendemboski Just curious, under what circumstances would an ember-electron project not already have package.json's `main` property set to `ember-electron/electron.js`?

jacobq:
Are you able to successfully run `ember electron` with this code? I can run `ember electron:package` and `ember electron:make` without problems, but running `ember electron` seems to cause an infinite loop because of a file watcher getting triggered.

I see something like this repeated continuously:
```
Slowest Nodes (totalTime => 5% )              | Total (avg)         
----------------------------------------------+---------------------
SourceMapConcat: Concat: App (1)              | 5ms                 
SourceMapConcat: Concat: Vendor /asset... (1) | 4ms                 
SourceMapConcat: Concat: Test Support JS (1)  | 3ms                 
SimpleConcatConcat: Concat: Addon CSS (1)     | 3ms                 
SourceMapConcat: Concat: Addon JS (1)         | 3ms                 
broccoli-persistent-filter:Babel (5)          | 2ms (0 ms)          
Funnel: Addon JS (8)                          | 2ms (0 ms)          
SourceMapConcat: Concat: App Tests (1)        | 2ms                 

file changed tmp/config_loader-input_base_path-c7YD29p2.tmp/0
file deleted tmp/broccoli_merge_trees-output_path-0xZTf2rS.tmp/package.json
file added tmp/broccoli_merge_trees-output_path-0xZTf2rS.tmp/package.json
file changed dist/package.json

Build successful - 33ms.



Slowest Nodes (totalTime => 5% )              | Total (avg)         
----------------------------------------------+---------------------
SimpleConcatConcat: Concat: Addon CSS (1)     | 3ms                 
SourceMapConcat: Concat: Test Support JS (1)  | 2ms                 
SourceMapConcat: Concat: App (1)              | 2ms                 
SourceMapConcat: Concat: Vendor /asset... (1) | 2ms                 
SourceMapConcat: Concat: Addon JS (1)         | 1ms                 
broccoli-persistent-filter:Babel (5)          | 1ms (0 ms)          
Funnel: Addon JS (8)                          | 1ms (0 ms)          
SourceMapConcat: Concat: App Tests (1)        | 1ms                 

file changed tmp/config_loader-input_base_path-c7YD29p2.tmp/0
file deleted tmp/broccoli_merge_trees-output_path-0xZTf2rS.tmp/package.json
file added tmp/broccoli_merge_trees-output_path-0xZTf2rS.tmp/package.json
file changed dist/package.json

Build successful - 37ms.
```

jacobq:
I tried adding the following guard just before the unlinking & re-writing of package.json:
```
// Only re-write package.json if we need to make a change
if (typeof json.main === 'string' && json.main.indexOf('electron.js') !== -1)
  return;
```

This seems to prevent the infinite loop but now a fatal error occurs while attempting to run the app (because it's looking for the app in the wrong path?):
```
jquant@fractal:~/Documents/KFMT/Software/ember-electron-update-demo$ ember electron

Build successful - 462ms.



Slowest Nodes (totalTime => 5% )              | Total (avg)         
----------------------------------------------+---------------------
Babel (16)                                    | 187ms (11 ms)       
Concat (8)                                    | 136ms (17 ms)       

Starting Electron...
App threw an error during load
Error: Cannot find module '/home/jquant/Documents/KFMT/Software/ember-electron-update-demo/dist'
    at Module._resolveFilename (module.js:470:15)
    at Function.Module._resolveFilename (/home/jquant/Documents/KFMT/Software/ember-electron-update-demo/node_modules/electron-prebuilt-compile/node_modules/electron/dist/resources/electron.asar/common/reset-search-paths.js:35:12)
    at Function.Module._load (module.js:418:25)
    at Module.require (module.js:498:17)
    at init (/home/jquant/Documents/KFMT/Software/ember-electron-update-demo/node_modules/electron-compile/lib/config-parser.js:291:16)
    at main (/home/jquant/Documents/KFMT/Software/ember-electron-update-demo/node_modules/electron-prebuilt-compile/lib/es6-init.js:40:29)
    at Object.<anonymous> (/home/jquant/Documents/KFMT/Software/ember-electron-update-demo/node_modules/electron-prebuilt-compile/lib/es6-init.js:43:1)
    at Object.<anonymous> (/home/jquant/Documents/KFMT/Software/ember-electron-update-demo/node_modules/electron-prebuilt-compile/lib/es6-init.js:45:3)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
A JavaScript error occurred in the main process
Uncaught Exception:
Error: Cannot find module '/home/jquant/Documents/KFMT/Software/ember-electron-update-demo/dist'
    at Module._resolveFilename (module.js:470:15)
    at Function.Module._resolveFilename (/home/jquant/Documents/KFMT/Software/ember-electron-update-demo/node_modules/electron-prebuilt-compile/node_modules/electron/dist/resources/electron.asar/common/reset-search-paths.js:35:12)
    at Function.Module._load (module.js:418:25)
    at Module.require (module.js:498:17)
    at init (/home/jquant/Documents/KFMT/Software/ember-electron-update-demo/node_modules/electron-compile/lib/config-parser.js:291:16)
    at main (/home/jquant/Documents/KFMT/Software/ember-electron-update-demo/node_modules/electron-prebuilt-compile/lib/es6-init.js:40:29)
    at Object.<anonymous> (/home/jquant/Documents/KFMT/Software/ember-electron-update-demo/node_modules/electron-prebuilt-compile/lib/es6-init.js:43:1)
    at Object.<anonymous> (/home/jquant/Documents/KFMT/Software/ember-electron-update-demo/node_modules/electron-prebuilt-compile/lib/es6-init.js:45:3)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
cleaning up...
jquant@fractal:~/Documents/KFMT/Software/ember-electron-update-demo$ 
```

jacobq:
In some cases the electron window would appear, but there was an error displayed that suggested an extra `dist/` getting appended somewhere.
```
Failed to compile /home/jquant/Documents/KFMT/Software/ember-electron-update-demo/dist/dist/index.html: ENOENT: no such file or directory, lstat '/home/jquant/Documents/KFMT/Software/ember-electron-update-demo/dist/dist'
Error: ENOENT: no such file or directory, lstat '/home/jquant/Documents/KFMT/Software/ember-electron-update-demo/dist/dist'
    at fs.lstatSync (fs.js:902:18)
    at Object.fs.lstatSync (ELECTRON_ASAR.js:244:16)
    at Object.realpathSync (fs.js:1542:21)
    at Object.fs.realpathSync (ELECTRON_ASAR.js:321:29)
    at cachedRealpath (/home/jquant/Documents/KFMT/Software/ember-electron-update-demo/node_modules/electron-compile/lib/sanitize-paths.js:25:22)
    at sanitizeFilePath (/home/jquant/Documents/KFMT/Software/ember-electron-update-demo/node_modules/electron-compile/lib/sanitize-paths.js:52:16)
    at FileChangedCache.getCacheEntryForPath (/home/jquant/Documents/KFMT/Software/ember-electron-update-demo/node_modules/electron-compile/lib/file-change-cache.js:201:48)
    at /home/jquant/Documents/KFMT/Software/ember-electron-update-demo/node_modules/electron-compile/lib/file-change-cache.js:127:41
    at Generator.next (<anonymous>)
    at step (/home/jquant/Documents/KFMT/Software/ember-electron-update-demo/node_modules/electron-compile/lib/file-change-cache.js:27:191)
```

index.js
bendemboski: This is the `package.json` at the root of the project. I removed the code from the blueprint that sets `main` to `ember-electron/electron.js`, so new apps would not have it set (although yours probably does because you ran the blueprint back when it was setting `main`).

The reason I did this is because I don't think setting `main` in the ember project's `package.json` really makes sense. We never run `ember-electron/electron.js` out of the Ember project, we "build" a new node package by copying the `package.json`, `ember-electron/electron.js`, the `public` files and the Ember build output to a new location and then run from there. So saying that the Ember project's main routing is `ember-electron/electron.js` doesn't make sense because we don't (and can't) actually run that file.

Moreover, I could imagine situations where an Ember application also does double-duty as a simple node package, and since a package only gets one `main` script, it seems best not to set it if we really don't need to.

bendemboski:
@jacobq I never saw the infinite loop, but a guard makes sense, although I think we could just check the `main` attribute and if it's already what we want, then no-op.

I'm not sure why your `ember electron` is failing -- can you post the output of `find dist/`?

jacobq:
@bendemboski part of that error was my fault. I updated my comment above because I hadn't re-run the blueprint, so my `electron.js` had the old path in it still. However, I do still get the infinite loop problem. The code I'm using to test is [here](https://github.com/jacobq/ember-electron-update-demo/), and I'll update this comment in a moment with the info your requested.

bendemboski:
@jacobq I just pushed a commit that should fix the infinite loop problem. Let me know if you're still seeing issues, and thanks!

index.js
jacobq: Ah, OK, I agree with your reasoning about not wanting to clutter a project's `package.json` with `ember-electron` configuration (whether that's the previously used `ember-electron` property with `copy-files` or the `main` property). In my case our apps using `ember-electron` don't have other deployment targets and aren't used by any other projects (shared parts are separated into their own modules). So it seems more complicated than necessary for those cases, but I can see where it might be useful in other situations.

jacobq:
I took your `broccoli-file-creator` idea and it seems to work. Is this what you were thinking?
https://github.com/jacobq/ember-electron/tree/broccoli-copy-files

jacobq:
It might also make sense to remove the `ember-electron` section from this project's `package.json` along with these changes, right?

bendemboski:
@jacobq oh yeah, missed that, thanks

index.js
```diff
@@ -82,6 +128,38 @@ module.exports = {
...
+          // to do!
+          return;
+        }
+
+        // On linux, package.json is a sym link at this point, so we need to
+        // unlink it before writing back to it, or else we'll end up modifying
+        // the file it points to, i.e. our project's package.json!
+        return unlink(packageJsonPath).then(() => {
+          json.main = path.join('electron', 'electron.js');
+          return writeJson(packageJsonPath, json, { spaces: 2 });
```
anulman: While we're at it, let's consider doing the same [for config](https://github.com/electron-userland/electron-forge#configuring-package)

```
json.config = json.config || {};
json.config.forge = "./resources/forge.json"
```

And replacing [`_ensurePackageJsonConfiguration`](https://github.com/felixrieseberg/ember-electron/blob/master/blueprints/ember-electron/index.js#L61-L85) with something that writes `packageJson.config.forge` to `ember-electron/forge.json`

pichfl:
@bendemboski I tried some things and probably made a whole mess in https://github.com/pichfl/ember-electron/tree/ember-schema-copy-files

But at least it should boot correctly when running `ember electron`. Maybe it's a starting point.

README.md
```diff
@@ -62,16 +62,29 @@ ember electron:package
...
+ │   ├── electron.js
+ │   └── <contents of public and public-<platform>
+ └── ember
+     └── <ember build output>
+```
+
+So, for example, when loading your ember app from `electron.js`, you want something like this (which is included in the default `electron.js` in the blueprint):
+
+```javascript
+const emberAppLocation = `file://${dirname}/../ember/index.html`;
```
pichfl: I think this might no longer match. I should totally add a short paragraph describing the usage of `electron-protocol-serve` here.

pichfl:
Do we need an ember-electron folder in the root? Shouldn't that be in /tests/dummy?

bendemboski:
@pichfl no, I don't think so. My understanding is that the `tests/dummy` directory itself isn't anything special, it's just that in Ember addons the three specific subdirectories it contains are set up to be the ones that are normally at the root in applications (see [here][(https://github.com/ember-cli/ember-cli/blob/e883ab84e6d255879272c04477a1c8e0c8e40867/lib/broccoli/ember-addon.js#L34)).

I don't know of a good way to set things up so that our `index.js` reads from the root of apps, but the dummy directory of addons when it's doing the broccoli pipeline stuff. Maybe it's possible, but would take some time to figure out.

PR #172, opened 2017-03-09
Removing ember-welcome-page

jacobq:
If I understand correctly, [`ember-welcome-page`](https://github.com/ember-cli/ember-welcome-page) is not necessary/used by addons, so I think it should be removed from this addon's `package.json`.


felixrieseberg:
Thanks!

Issue #176, opened 2017-03-12
Beta: ember-welcome-page


pichfl:
Extracted from #173 

Up to now I can report this:

1. ember-welcome-page uses a template tag `welcome-page` and exports assets now. No more express middleware.
2. This page loads fine when running `ember electron`
3. It does not load when running a packaged app packaged with `--env=production`, which is our default, this happens, because the addon is configured to bail if in production. [See code here](https://github.com/ember-cli/ember-welcome-page/blob/master/index.js#L10).
4. When running an app packaged with `--env=development`, it works as well.

We should consider:

1. [x] Warning when `ember-welcome-page` is detected
2. [x] Describe the behaviour in our README or docs
3. Create our own `ember-electron-welcome` as suggested by @bendemboski 

While I think 3) is quite interesting, I don't think it is that useful. While some people might start using our addon to create a brand new app, coming from the ember side with a working app in place is a equally likely scenario. I don't want to mess too much with other peoples apps, especially their templates. Therefore putting out a warning if we detect both the template tag and the addon (or just the addon) we should put out a big warning. If we detect both, we might be able to assume the user runs `ember-electron` from a brand new app and replace the addon with our own page, but that seems like a lot of work right now. Showing a link to a possible "Getting Started with ember-electron" next to the warning in the console should be sufficient.

anulman:
@pichfl I like the idea of warning when ember-welcome-page is detected, at least for all the `commands/-build` variants.

Can you take this on, or otherwise close this issue given the docs branch is merged?

pichfl:
Will take care of this once #188 lands.

pichfl:
Closed in favor of #203

PR #178, opened 2017-03-13
WIP: Move documentation to /docs

pichfl:
This should provide a little more structure and allow us to generate a proper website some day in the future.

This will also help to resolve issues like #173 and #176


felixrieseberg:
Good idea!

PR #181, opened 2017-03-13
Add electron-out to .npmignore

bendemboski:
Don't want to include it in release packages if it's present!

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -2,6 +2,7 @@ bower_components/
 tests/
 tmp/
 dist/
+electron-out/
 
 .bowerrc
 .editorconfig


jacobq:
Would it make sense to also do something like this to an incorporating app's `.npmignore`? (i.e. like we do in the blueprint in `_updateGitignore`)

bendemboski:
I suppose that could make sense, although it's kind of an edge case. That file only matters when you're packing an application to publish to npm, which isn't really a use case for Ember applications. It is for addons, so the edge case we could support is installing `ember-electron` into an addon. So I guess it wouldn't hurt to add logic to do that, but you'd definitely have to check to make sure `.npmignore` exists, and you wouldn't want to create it if it doesn't (because then you're not running in a project that's meant to be packed, or the project just relies on the `.gitignore` file).

I don't see it as a very high priority, but if you want to make it happen, go for it 😄

anulman:
@bendemboski @jacobq let's separate those PRs. Merging this

Issue #154, opened 2017-03-03
TypeError: Path must be a string. Received undefined


pichfl:
I'm really curious about this piece in `ember-electron/electron.js`

```js
const dirname = __dirname || path.resolve(path.dirname());
```

`__dirname` is a local to each module and should never be empty.

`path.dirname()` however will always throw because it expects a string input:

```sh
> path.dirname()
TypeError: Path must be a string. Received undefined
    at assertPath (path.js:7:11)
    at Object.dirname (path.js:1324:5)
    at repl:1:6
    at ContextifyScript.Script.runInThisContext (vm.js:23:33)
    at REPLServer.defaultEval (repl.js:334:29)
    at bound (domain.js:280:14)
    at REPLServer.runBound [as eval] (domain.js:293:12)
    at REPLServer.onLine (repl.js:531:10)
    at emitOne (events.js:101:20)
    at REPLServer.emit (events.js:189:7)
```

Is there a reason why this is here? I think it was introduced in https://github.com/felixrieseberg/ember-electron/commit/340ac2657df211c223b4e5aca9bb68bca5c4f480 and should not have worked back there either? I think this was introduced because of #56 but the answers on the linked stack overflow was never quite properly used (missing an empty string). 

I think it had something to do with very old versions of Ember-CLI. Do we still support CLI before 2.0?

If so, we should fix it with 
```js
const dirname = __dirname || path.resolve(path.dirname(''));
```
or we should remove this completely.

anulman:
Good hunting @pichfl! I've not audited for e-cli 1.x support in a while, but don't remember a decision to _drop_ support for the last gen.

I'd prefer the latter solution for now and to drop 1.x support in a later PR (preferably when we start supporting 3 😆)

kevingelion:
I've definitely seen this before and was never on e-cli 1.x. Could this be related to an older node version? I think I accidentally tried installing `e-electron` on a project running node `4.6.2`.

pichfl:
Since I'm already working on the blueprints in #155 and #156 I'm pretty sure I can sneak that fix in there.

pichfl:
Done. See #155.

jacobq:
[...]

Issue #56, opened 2016-04-23
__dirname is not defined in blueprint template


kevinmickey:
Same issue brought up in by gordonbisnor in #36, but it seems unrelated to the original issue that was closed, so I'm opening this issue.

https://stackoverflow.com/questions/8817423/node-dirname-not-defined

I've tried using node v5.2.0 and v5.11.0.

I run ember install ember-electron and get:

>   [?] Overwrite config/environment.js? No, skip
  skip config/environment.js
  create electron.js
  create tests/electron.js
  create tests/package.json
__dirname is not defined (Error in blueprint template: /Users/kevin/Web/mycms/netlify-cms/node_modules/ember-electron/blueprints/ember-electron/files/electron.js)
ReferenceError: __dirname is not defined (Error in blueprint template: /Users/kevin/Web/mycms/netlify-cms/node_modules/ember-electron/blueprints/ember-electron/files/electron.js)
    at eval (eval at <anonymous> (/Users/kevin/Web/mycms/netlify-cms/node_modules/ember-cli/node_modules/lodash/string/template.js:214:12), <anonymous>:8:10)
    at processTemplate (/Users/kevin/Web/mycms/netlify-cms/node_modules/ember-cli/lib/models/file-info.js:13:52)
    at /Users/kevin/Web/mycms/netlify-cms/node_modules/ember-cli/lib/models/file-info.js:94:20
    at lib$rsvp$$internal$$tryCatch (/Users/kevin/Web/mycms/netlify-cms/node_modules/rsvp/dist/rsvp.js:493:16)
    at lib$rsvp$$internal$$invokeCallback (/Users/kevin/Web/mycms/netlify-cms/node_modules/rsvp/dist/rsvp.js:505:17)
    at lib$rsvp$$internal$$publish (/Users/kevin/Web/mycms/netlify-cms/node_modules/rsvp/dist/rsvp.js:476:11)
    at lib$rsvp$asap$$flush (/Users/kevin/Web/mycms/netlify-cms/node_modules/rsvp/dist/rsvp.js:1198:9)
    at _combinedTickCallback (internal/process/next_tick.js:67:7)
    at process._tickCallback (internal/process/next_tick.js:98:9)

timbaas:
I'm experiencing the same issue on node v4.2.6

JoshuaTyree:
[...]

PR #183, opened 2017-03-14
Replace path.resolve(path.dirname()) -> process.cwd()

jacobq:
This may really be just a cosmetic thing, but as was noted in #154, there appears to be some broken code regarding a work-around for missing `__dirname` (see also #56). It looks like there are several places where we have code like `__dirname || path.resolve(path.dirname())`.

As @pichfl noted, [`path.dirname()` throws `TypeError: Path must be a string. Received undefined`](https://github.com/nodejs/node/blob/98e54b0bd4854bdb3e2949d1b6b20d6777fb7cde/lib/path.js#L1345), so it shouldn't be called without arguments. Secondly, I'm pretty sure this is extraneous code to begin with since calling `path.resolve()` (no arguments) returns a string containing the CWD, which I think is the same as `path.resolve('.')` (which should be the same as `path.resolve(path.dirname(''))`).

The [node v4.x docs](https://nodejs.org/dist/latest-v4.x/docs/api/path.html#path_path_resolve_from_to) weren't particularly clear about this:
> Note: If the arguments to resolve have zero-length strings then the current working directory will be used instead of them.

but the [latest ones](https://nodejs.org/docs/latest/api/path.html) make this explicit:
> If no path segments are passed, path.resolve() will return the absolute path of the current working directory.

and a quick, manual test in the 4.x REPL suggest the behavior has not changed.

Is there any reason we wouldn't want to remove this extra (potentially typo-prone) call?

### DIFF
--- a/app/electron/reload.js
+++ b/app/electron/reload.js
@@ -16,7 +16,7 @@
    * @param sub directory
    */
   let watch = function(sub) {
-    let dirname = __dirname || path.resolve(path.dirname());
+    let dirname = __dirname || process.cwd();
     let isInTest = !!window.QUnit;
 
     if (isInTest) {
@@ -74,7 +74,7 @@
   };
 
   document.addEventListener('DOMContentLoaded', (/* e */) => {
-    let dirname = __dirname || path.resolve(path.dirname());
+    let dirname = __dirname || process.cwd();
 
     fs.stat(dirname, (err/* , stat */) => {
       if (!err) {
--- a/index.js
+++ b/index.js
@@ -6,7 +6,7 @@ let path = require('path');
 let { clone } = require('lodash/lang');
 
 function injectScript(scriptName) {
-  let dirname = __dirname || path.resolve(path.dirname());
+  let dirname = __dirname || process.cwd();
   let filePath = path.join(dirname, 'lib', 'resources', scriptName);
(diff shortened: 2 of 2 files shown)


jacobq:
Also, @anulman mentioned using `process.cwd()` instead of `path.resolve()` to be more clear. I think I'll update this and re-push.

pichfl:
Can't we actually remove __dirname all together and just go with `process.cwd()` all together?

jacobq:
@pichfl I'm not sure if that would break anything, so I went for the more conservative change. (Once the test suite is back online I will feel much better.) I think there is a slight difference though, as 
[`__dirname`](https://nodejs.org/api/globals.html#globals_dirname) is different depending on where a module is located, I think.

felixrieseberg:
I'm down for this change. Ideally, we should never hit a scenario in which `__dirname` is undefined, but you know, computers are weird!

jacobq:
One thing we may want to be careful of here is that it seems `process` is not always defined in the electron renderer process. I didn't see this in [the documentation](https://electron.atom.io/docs/api/process/) but, for example, in ee's `reload.js` it's undefined in the main IIFC/closure but it seems fine once the `DOMContentLoaded` event fires.

pichfl:
Did you check for window.processNode? Our shim removes window.process in <head> and reinstates it only later before </body>.

PR #186, opened 2017-03-15
Watch `dist/ember` instead of `__dirname` in `reload.js` (to fix #182)

### LINKED ISSUE #182: Beta: Uncaught Error: watch [...]/electron.asar/renderer ENOTDIR
This error shows up when running `ember electron` as well as when running an app packaged with `ember electron:package`.

```
Uncaught Error: watch /Users/pichfl/Repositories/tron/node_modules/electron-prebuilt-compile/node_modules/electron/dist/Electron.app/Contents/Resources/electron.asar/renderer ENOTDIR
```

jacobq:
It seems that [`reload.js` was attempting to use `__dirname`](https://github.com/felixrieseberg/ember-electron/blob/master/app/electron/reload.js#L19) to locate the front-end / ember app files, and somewhere along the way this stopped working (perhaps when switching to `electron-forge` or changing the structure of the `dist` output folder). (See #182) Although there may be a better solution than this, I just changed `reload.js` to use `path.resolve('dist', 'ember')` instead since this appears to do the trick.

If anyone has comments / feedback / alternatives / improvements feel free to let me know, and I'll be happy to update this with something better.

### DIFF
--- a/app/electron/reload.js
+++ b/app/electron/reload.js
@@ -7,6 +7,10 @@
   // Reload the page when anything in `dist` changes
   let fs = window.requireNode('fs');
   let path = window.requireNode('path');
+  // Note: This path assumes we're running in development, but
+  // if app.env !== 'development' then this file shouldn't have been imported at all
+  // (see ../../index.js)
+  const emberDistDir = path.resolve('dist', 'ember');
 
   /**
    * @private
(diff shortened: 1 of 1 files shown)


anulman:
Thanks for this @jacobq. Fixed along with several other items in #188

Issue #175, opened 2017-03-12
Uncaught Error: Cannot find module 'devtron'


pichfl:
When running the packaged app and opening the dev tools (made available via code), it can't seem to find the devtron module.

It should be fixed by adding devtron as a dependency, but maybe we should make our installation of devtron a bit more stable, only installing if the dependency was found and is included?

jacobq:
I see this in the case of `ember electron` but not `ember electron:package`. Is that the same as your experience? Running `npm install --save devtron` does not seem to help.

jacobq:
Also, it seems that even though the error message appears in the case of `ember electron`, the Devtron 'tab' is there whereas in packaged builds there is neither an error nor the corresponding tab. Furthermore, attempting to run `require('devtron')` in a packaged version causes the error to appear again. Seems related to `resources/electron.asar/common/reset-search-paths.js`

pichfl:
Nope, happens to me in `ember electron:package` as well. Or well, scratch that, I had not installed devtron as a dependency at all. Strangely enough, devtron seems to be missing only when running in `--env=development` (which is the case for `ember electron` but not for `ember electron:package` by default. Adding the flag manually shows the error in both builds.

jacobq:
Hacky but appears to work:
```
const fs = window.requireNode('fs');
const path = window.requireNode('path');
const module = window.requireNode('module');

// Check if we're running with resource root inside of a `electron-prebuilt-compile` dir.
// This should correspond to when running `ember electron`
const epcIndex = __dirname.indexOf('electron-prebuilt-compile');
if (epcIndex !== -1) {
  //let epcNodeModulesPath = path.resolve(__dirname.slice(0, epcIndex), '..', 'node_modules');
  const epcNodeModulesPath = __dirname.slice(0, epcIndex);
  if (fs.statSync(epcNodeModulesPath).isDirectory()) {
    module.Module.globalPaths.push(epcNodeModulesPath);
  }
}

// When running a packaged app, we have our modules in
// resources/app/node_modules
const eeNodeModulesPath = path.resolve(process.resourcesPath, 'app', 'node_modules');
if (fs.statSync(eeNodeModulesPath).isDirectory())
  module.Module.globalPaths.push(eeNodeModulesPath);
```

I think this is partly an upstream problem with `electron-prebuilt-compile`

bendemboski:
[...]

PR #187, opened 2017-03-15
Work-around bad module path restrictions

jacobq:
There seems to be an [upstream problem](https://github.com/electron-userland/electron-prebuilt-compile/issues/23) (or "feature" -- haven't figure it out completely yet) that prevents front-end code from being able to find modules (i.e. call `require` without path). For example, as described in #175, when running `ember electron` the `devtron` module won't load using `require('devtron')` even though the module is present in `<project>/node_modules`. 

Hopefully something will be done to address this upstream (if that's indeed the cause of the this), but in the mean time this patch adds either the project root's `node_modules` folder (i.e. when running `ember electron`) or the one inside electron's resource folder under `app/node_modules`. Although I'm not sure if this is a complete or proper solution, it seems to work and may be useful to others. Comments / feedback welcome.

### DIFF
--- a/app/electron/adjust-module-search-paths.js
+++ b/app/electron/adjust-module-search-paths.js
@@ -7,17 +7,19 @@
   var path = window.requireNode('path');
   var module = window.requireNode('module');
 
-  var pathsToAdd = [
-    path.join(window.processNode.cwd(), 'node_modules'),
-    path.resolve(window.processNode.resourcesPath, 'app', 'node_modules')
-  ];
+  var searchPaths = window.processNode.mainModule.paths;
+  var packagedNodeModulesPath = path.resolve(window.processNode.resourcesPath, 'app', 'node_modules');
+  var projectRootNodeModulesPath = (function() {
+    // n.b. if we are using `electron-prebuilt-compile` seek top-level node_modules dir
+    var epcIndex = __dirname.indexOf('electron-prebuilt-compile');
+    if (epcIndex !== -1) {
+      return __dirname.slice(0, epcIndex);
+    }
+  }());
 
-  // n.b. if we are running in `electron-prebuilt-compile`,
-  // TODO @jacobq is this still necessary given window.processNoce.cwd() above?
-  //      - if so, plz explain what __dirname is, why you're slicing, etc
-  var epcIndex = __dirname.indexOf('electron-prebuilt-compile');
-  if (epcIndex !== -1) {
-    pathsToAdd.push(__dirname.slice(0, epcIndex));
+  var pathsToAdd = [packagedNodeModulesPath];
+  if (projectRootNodeModulesPath) {
+    pathsToAdd.push(projectRootNodeModulesPath);
   }
 
   pathsToAdd.forEach(function(pathToAdd) {
@@ -25,12 +27,11 @@
       // n.b. use statSync to prevent race conditions
       var stats = fs.statSync(pathToAdd);
 
-      // TODO @jacobq explain why you are setting globalPaths vs paths
-      if (stats.isDirectory() && module.globalPaths.indexOf(pathToAdd) === -1) {
-        module.globalPaths.push(pathToAdd);
+      if (stats.isDirectory() && searchPaths.indexOf(pathToAdd) === -1) {
+        searchPaths.push(pathToAdd);
       }
     } catch(e) {
-      // n.b. ignore "dir does not exist" errors, else throw
+      // n.b. ignore "dir does not exist" errors (app may not be packaged)
       if (e.code !== 'ENOENT') {
         throw e;
       }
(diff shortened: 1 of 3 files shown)


anulman:
@jacobq the devtron-related issue is resolved in #188. Are there other examples of libs you're having trouble requiring in front-end / Ember code without this patch? I was unable to replicate this otherwise.

jacobq:
If you have another solution for the `devtron` issue then that'll probably be (1) better and (2) potentially applicable to other cases. I can't think of anything off-hand (my code references some private modules but does it using a path). Plus, I can always pull this hack out of my back pocket if I need it, so I'll close this.

bendemboski:
Actually, I think this is still an issue. I have a third-party module (`tmp`) installed in my `dependencies` and I cannot import it from my running application under `ember electron` or `ember electron:test`. Why do we think this is not longer an issue?

jacobq:
@bendemboski I haven't tried the latest work yet, so there may very well still be an issue, but my thinking was that this:
* The primary motivation is to fix the devtron error which I believe the other work does
* My search path hack is probably the WrongWay™ to fix this
* In my application code I can work-around this by either explicitly specifying a path to my modules or stuffing this hack if I want to
* I did not know anyone else had a use case / problem so didn't want to be offering a "solution in search of a problem"

bendemboski:
Yeah, I just ran into the problem myself. Do you have any npm modules that aren't node core (i.e. that you installed in your `dependencies` in `package.json`) that you try to access from a renderer process? It looks to me like that's broken somehow, and it's a pretty core use case IMO.

jacobq:
Right, if you try to access a non-built-in module from the renderer process w/o specifying the path then it won't work. You're boosting my confidence that this truly is/was a bug and that I'm not completely off in la la land :) Perhaps most noteworthy is that this was not the case before switching to `electron-forge` and it is similar to other bug reports people are filing upstream, so if that's really where the problem is then I think that'd be the best place to address it. I'll re-open this PR for your consideration, but it is, at best, a band-aid we'll probably have/want to take out later.

bendemboski:
Actually, I'm not pretty sure this only happens when using `electron-protocol-serve`. It looks like maybe Electron or somebody can't find `index.html` to set it as the main module's path, so it gets set to `node_modules/electron/dist/Electron.app/Contents/Resources/electron.asar/renderer/init.js`.

@pichfl any thoughts here?

anulman:
This was very helpful.

@jacobq I'm going to push a commit to your branch with the fixes I needed to get this running locally, as well as a bunch of text rm'ed. Before landing, can you please address the two todos I left?

I should also add that several of your notes may be worth adding back. I removed them for now because comments like these tend to get out of sync with the code they're commenting on, which can be pretty confusing on subsequent reads. Can you try to keep in-code comments minimal & terse, and otherwise document necessary knowledge in the README (or now @pichfl's docs dir)?

bendemboski:
This is actually a big problem, and one that I think we should figure out how to solve directly. As I mentioned in https://github.com/felixrieseberg/ember-electron/issues/175#issuecomment-286964364, the problem is that using `electron-protocol-serve` to load `index.html` via a `serve://` URI prevents [this](https://github.com/electron/electron/blob/5200b1a1ab6f76f5944c6d5f728bc51827101b31/lib/renderer/init.js#L106) code from setting up the renderer process' main module to be in the right directory. This kinda breaks a fundamental assumption about how `electron` works, and I think we need to figure out how to solve it directly, rather than work around it by guessing at the right values to go in the global module search path.

I'll put some thought into it, but my initial idea is to load `index.html` via a `file:` URI like we normally do, but have our build pipeline inject a `<base>` tag into `index.html` to make all the other assets use `serve://` URLs.

anulman:
Ah! Yeah, that'd do it. Didn't see the link in your prior comment.

pichfl:
@bendemboski I still have to play catch up here, but my protocol might indeed be the problem.

Except that loading the index.html via file would not solve the other problem the protocol solves: The fact that file urls have vastly different content-security-policies, what messes with many libraries like three.js etc.

Also I was very happy to see Ember dropping the `<base>` tag as it interferes with linked svg images. 

So the solution proposed in our channel, returning an environment-variable from the protocol (or setting one where the protocol is set up) seems like a good option to me that we could use in combination with our shims.

jacobq:
I think this should be closed in favor of #193. Thoughts?

bendemboski:
Yeah, I'm pretty sure #193 hits the root cause of this. But thanks a lot @jacobq for the investigation that led back to the module include paths -- that did not even occur to me until I saw this PR 😄

jacobq:
You're welcome; I'm glad I could help. I'm going to actually try running your fix now. Once that lands let's close this.

PR #188, opened 2017-03-15
Use forge import & update file structure

anulman:
This PR affects blueprints & build output. Specifically, it:

- Uses upstream electron-forge's `api/import`, so users' projects conform to forge expectations;
- Makes `ember` & `ember-electron` sibling dirs in built output, to better match built dir structures to those of your working dir; and
- Moves `ember-electron/index.js` out of the `lib` dir, to keep it as high-visibility as forge will allow (i.e. "at least one dir away from package.json")

A dir structure of 
```
├── package.json
├── app
│   ├── <ember app files>
├── ember-electron
    ├── .compilerc
    ├── .electron-forge
    ├── index.js
    ├── resources
    │   ├── <files>
    ├── resources-darwin
    │   ├── <files>
    ├── resources-linux
    │   ├── <files>
    ├── resources-win
           ├── <files>
```

Gets built as
```
├── package.json
├── .compilerc
├── ember
│   ├── <ember build output>
├── ember-electron
    ├── .electron-forge
    ├── index.js
    ├── resources
        ├── <file copied/merged from resources and resources-<platform>>
        ├── <file copied/merged from resources and resources-<platform>>
```

Builds are output to `electron-out`.


app/electron/reload.js
```diff
@@ -38,7 +38,8 @@
    * Install Devtron in the current window.
    */
   let installDevtron = function() {
-    let devtron = window.requireNode('devtron');
+    let pathToDevtron = window.processNode.cwd() + '/node_modules/devtron';
```
bendemboski: Will this work on windows? Mixing `\` with `/`?

blueprints/ember-electron/index.js
```diff
@@ -24,96 +21,63 @@ class EmberElectronBlueprint extends Blueprint {
...
-      packages: ['electron-protocol-serve@1.1.0', ...deps]
-    }));
+  _installElectronTooling(logger) {
+    logger.startProgress('Installing electron build tools');
+
+    return efImport({
+      updateScripts: false,
+      outDir: 'electron-out',
+    })
+      .then(() => this.addPackageToProject('devtron', '^1.4.0'))
```
bendemboski: Since we're already using the task below, I think it would be more readable to use the task here as well instead of the `addPackageToProject` helper. That makes it clearer to the reader the difference between how we are installing `devtron` and how we are installing `electron-protocol-serve`.

app/electron/reload.js
anulman: Derp, good catch. We have access to path here

blueprints/ember-electron/index.js
```diff
@@ -24,96 +21,63 @@ class EmberElectronBlueprint extends Blueprint {
...
-    });
+  _createResourcesDirectories(logger) {
+    const ensureFile = denodeify(fs.ensureFile);
 
-    return all(promises);
+    logger.startProgress('Creating ember-electron resource dirs');
+
+    let rootDir = this.options.project.root;
+    let promises = [null, 'darwin', 'linux', 'win32']
+      .map((platform) => platform ? `resources-${platform}` : 'resources')
```
bendemboski: To my eye, this step makes it harder to read because I have to think to figure out what directories we are creating. Why not just `['resources', 'resources-darwin', 'resources-linux', 'resources-win32'].map()`?

blueprints/ember-electron/index.js
anulman: I was on the fence about this too, but ultimately opted for the "use available upstream tooling" option (again, since if we're working around it we ought to focus on landing it upstream instead).

I think the brief `// n.b. ...` note below explains why the change in form, and would prefer leaving this as is. Happy to wait on another voice here if you consider this strong dissent / a blocker.

blueprints/ember-electron/index.js
bendemboski: And why the two `path.join()` steps below? Couldn't this all be a very simple `forEach()` loop?

anulman: Haha, works for me

blueprints/ember-electron/index.js
```diff
@@ -24,96 +21,63 @@ class EmberElectronBlueprint extends Blueprint {
...
+    const writeJson = denodeify(fs.writeJson);
+
     let packageJsonPath = path.join(this.project.root, 'package.json');
-    // We can't use this.project.pkg because we modified package.json by
-    // installing dependencies, and that won't be reflected in this.project.pkg
-    return readJson(packageJsonPath)
-      .then((packageJson) => {
-        packageJson.config = packageJson.config || {};
-        packageJson.config.forge = './ember-electron/.electron-forge';
+    let forgeConfigPath = './ember-electron/.electron-forge.js';
```
bendemboski: Why `.js`? I though we were trying to make this look kinda like `.ember-cli` here, which doesn't have a `.js` extension?

blueprints/ember-electron/index.js
```diff
@@ -24,96 +21,63 @@ class EmberElectronBlueprint extends Blueprint {
...
-        } else {
-          lines.push('');
-          lines.push(toAdd);
-        }
-
-        return writeFile(gitignorePath, lines.join('\n'));
-      });
+    return readJson(packageJsonPath)
+      .then((packageJson) => {
+        let forgeConfig = `module.exports = ${packageJson.config.forge}`;
```
bendemboski: Isn't this going to end up being `module.exports = [object Object]`?

blueprints/ember-electron/index.js
pichfl: I think having an extension here as done by eslint and others is really helpful and makes configuring editors much easier.

blueprints/ember-electron/index.js
bendemboski: The tasks are just as public of an API as the blueprint method in ember-cli, so we're not using anything private. If things change such that using the task directly breaks (which is unlikely since it's a public API), then installing the procotol handler will break anyway, so I don't think we're saving ourselves anything here, just making the code look more confusing. Also, the `addPackageToProject` was designed for Ember projects where of course everything is a `devDependency` because it's all built via the toolchain so it really doesn't matter, so why not make it a `devDependency`. Using that method here hides the fact that it's important to us that it is explicitly `devDependency` and not a "whatever kind of dependency the blueprint class chooses to default to".

Doesn't matter at runtime, but I think code readability, especially in an open source project, should be a priority.

blueprints/ember-electron/index.js
bendemboski: Okay, I don't have a strong opinion, I'm just pointing out that one of our stated goals in moving it to the file where it is now was to make it more like the hopefully-familiar `.ember-cli`, which does not have an extension.

anulman: Fair enough, we don't need it. In the spirit of your point though, `.ember-cli` is a json file (I recognize we're using .js to make platform-specific changes a little easier to config)

blueprints/ember-electron/index.js
anulman: Fair enough. Don't feel strongly either way, will make this change too 👍

blueprints/ember-electron/index.js
```diff
@@ -24,96 +21,79 @@ class EmberElectronBlueprint extends Blueprint {
   afterInstall(/* options */) {
     let logger = new Logger(this);
 
-    return this._installDependencies(logger)
-      .then(() => this._createResourcesDirectories())
-      .then(() => this._ensurePackageJsonConfiguration())
-      .then(() => this._updateGitignore())
```
bendemboski: Don't we still need to update `.gitignore` to include `electron-out`? Or does forge take care of that for us? In any case, in #189 we'll need to add `electron-livereload`.

anulman: Waiting to hear back from upstream. But yes, we ought to handle this :)

anulman: To be fair, I don't think this should be considered blocking, as we're tracking in #189 & a lot of subsequent work will likely be contingent on this.

bendemboski: Okay, I guess we can fix it in another PR, but it seems weird to delete the code that does what we want to do in one PR just to re-add it in another PR...

anulman: Fine 🙄 😛

bendemboski:
@anulman would you mind updating the description of this PR? I don't think it fixes the issues you mention, and it would be nice to have a summary of what it does and why for posterity's sake (so when we're searching back through the commit log trying to figure out why we did something, we can see a brief explanation).

anulman:
@bendemboski @pichfl @felixrieseberg updated the description, + the structure docs

blueprints/ember-electron/index.js
```diff
@@ -24,96 +21,143 @@ class EmberElectronBlueprint extends Blueprint {
...
+    ]
+      .map((dirName) => path.join(rootDir, 'ember-electron', dirName))
+      .map((dirPath) => ensureFile(path.join(dirPath, '.gitkeep')));
+
+    return all(ensureResourceDirGitKeeps)
+      .then(() => logger.message('Created ember-electron resource dirs'));
   }
 
-  _ensurePackageJsonConfiguration() {
+  _ensureForgeConfig(logger) {
```
bendemboski: Can you include a few comments in the function to explain why we are bailing in some of the early-out cases? Either a comment at the top explaining the whole plan about when we do/don't detect forge configs and when it's inline in package.json vs. in an external file, or else specific comments by each early-out `return` explaining why we are bailing in that case? I'm just worried that all that logic will easily slip out of our brains in the future, and will be nice to have comments reminding us.

blueprints/ember-electron/index.js
```diff
@@ -24,96 +21,143 @@ class EmberElectronBlueprint extends Blueprint {
...
+
+        delete packageJson.config.forge;
+
+        if (Object.keys(packageJson.config).length === 0) {
+          delete packageJson.config;
+        }
 
-        return writeJson(packageJsonPath, packageJson, { spaces: 2 });
+        return all([
+          writeFile(forgeConfigPath, `module.exports = ${forgeConfig}`),
```
bendemboski: We should end the file content with a `;` to keep consisting with generally accepted default linting standards.

docs/faq/structure.md
```diff
@@ -1,32 +1,45 @@
...
+The `ember-electron` folder created at the root of your project by
+`ember-electron`'s blueprint looks like this:
 
 ```
  ember-electron
  ├── .compilerc
  ├── .electron-forge
- ├── lib
- │   └── index.js
+ ├── index.js
```
bendemboski: Should be `main.js` right?

docs/faq/structure.md
```diff
@@ -1,32 +1,45 @@
...
- ├── resources
- └── ember
-     ├── index.html
-     └── <the rest of your built ember app>
+ ├── .compilerc
+ ├── ember
+ │   ├── <ember build output>
+ ├── ember-electron
+ │   ├── .electron-forge
+ │   ├── index.js
```
bendemboski: Also `main.js` here

ember-electron/main.js
```diff
@@ -4,6 +4,9 @@ const protocolServe = require('electron-protocol-serve');
 
 let mainWindow = null;
 
+console.log(__dirname);
```
bendemboski: Left in some debugging logs

index.js
```diff
@@ -49,63 +49,72 @@ module.exports = {
...
-      packageJson.config.forge = './.electron-forge';
+      packageJson.config.forge = packageJson.config.forge
+        || 'ember-electron/.electron-forge';
 
       let trees = [
-        // write package.json
         writeFile('package.json', JSON.stringify(packageJson, null, '  ')),
-        // copy the rest of the ember-electron directory, with the exception of
-        // the resources directories that need to be merged
+        writeFile('.compilerc', fs.readFileSync(compileRcPath)),
```
bendemboski: Your input and output here are both files, so rather than using `writeFile()` you should just a funnel:

```javascript
funnel(emberElectronPath, {
  files: ['.compilerc']
})
```

lib/commands/build.js
```diff
@@ -133,7 +133,7 @@ module.exports = Command.extend({
 
     return copy(
       path.resolve(this.project.root, distPath),
-      tmpPath
+      path.join(tmpPath, 'ember')
```
bendemboski: Don't you need to make this change to `_buildAppToTmp()` below as well?

blueprints/ember-electron/index.js
anulman: I really don't like comments inline, as they get out of sync quickly. This documentation belongs on the site.

blueprints/ember-electron/index.js
anulman: Good catch

docs/faq/structure.md
anulman: Yup! Fixing.

docs/faq/structure.md
anulman: Agreed, see note about these comments being easy to ignore / go out of sync.

ember-electron/main.js
anulman: Derp!

index.js
anulman: 👍

blueprints/ember-electron/index.js
bendemboski: Then maybe mention that? Something like "this implements the policy on how we treat pre-existing electron forge configs -- see addon documentation for more info"? I just look at this and see a bunch of cryptic code that without contextual knowledge is baffling, so at least a comment saying that an explanation exists somewhere else would be helpful!

docs/faq/structure.md
bendemboski: Yeah, I'd be fine with killing this whole comment and just pointing readers to the README or documentation site or whatever if you'd rather do that than update this comment.

lib/commands/build.js
anulman: All of this should have been deprecated. Weird.

blueprints/ember-electron/index.js
anulman: Will be removing this again, as outDir is supported upstream + we have bo other ignores.

tests/ember-electron/main.js
```diff
@@ -47,15 +47,3 @@ app.on('ready', function onReady() {
     mainWindow = null;
   });
 });
-
-let cwd = __dirname || resolve(dirname());
```
bendemboski: Uh wait, why did you remove this? IMO it's an important part of our "integration" testing that verifies that we are assembling the broccoli pipeline correctly.

blueprints/ember-electron/index.js
bendemboski: Sweet :)

blueprints/ember-electron/index.js
anulman: Let’s please have this discussion in an issue or a PR to the gh-pages branch. It will be much easier to make the right decision when we have a sense of the remaining docs, and I don’t believe this is a blocker functionally.

I am otherwise working on a test branch that will resolve this concern by documenting in spec.

tests/ember-electron/main.js
anulman: I rm’ed it because I ran ember g ember-electron and it was rm’ed. IIRC this behaviour is tested in the unit tests; if it isn’t it ought to live there vs. generated code.

bendemboski: I agree that it doesn't belong in the blueprint, but the whole `tests/` directory of this project is basically an integration test for the project itself. It allows us to run `ember electron:test` and make sure that various things work. So, while we wouldn't want this code in the blueprint because it doesn't belong in applications using this addon, it makes a lot of sense to include it here so that every time we run `ember electron:test` in *this* project we are also verifying that we can access native dependencies.

I don't think it makes sense to try to require that the test code that we have in our dummy test app in this addon exactly matches what's in the blueprint.

anulman: How is this different from what lives here: https://github.com/felixrieseberg/ember-electron/blob/master/tests/unit/end-to-end-test.js#L18-L22

I'm not so much opposed, as I think there's better form for this and we ought to be focusing on the testing branch (+ tbh probably gitignoring the generated `ember-electron` & `tests` dirs).

bendemboski: This is in the main process, and the other is in the renderer process, and because of the workaround for module include paths in the render process' shim, they are fairly different setups that I think need to be tested independently.

I entirely agree that we could probably have better and more thorough ways of testing this main-process functionality and ensuring that the package is assembled correctly and whatnot, but this is what we have right now, and it's been really really helpful to me as a smoke test when making changes to the addon, so I'd really like to keep it until we have something that replaces it.

index.js
```diff
@@ -49,63 +49,75 @@ module.exports = {
...
 
-      // Set the main entry point for the electron application, and point the
-      // electron-forge config to ./.electron-forge
       let packageJson = clone(this.project.pkg);
-      packageJson.main = 'lib/index.js';
+      let platform = process.env.EMBER_CLI_ELECTRON_BUILD_PLATFORM
+        || process.platform;
+
+      let emberElectronPath = path.join(this.project.root, 'ember-electron');
+      let compileRcPath = path.join(emberElectronPath, '.compilerc');
```
pichfl: This is unused. Do we still need this somewhere?

anulman: Compilerc is passed to `electron-compile`, and is a required part of forge's upstream packaging flow.

pichfl: Yes, but the variable is not in use at all. This could be a merge artefact?

PR #189, opened 2017-03-16
refactor(electron-start): build & cleanup to `electron-livereload` dir

anulman:
Rationale: let's not overwrite `dist` with unfamiliar build artifacts that may get packaged up and deployed somewhere by unwitting scripts.

Probably best to land this after #188.


anulman:
Includes work from #187

PR #191, opened 2017-03-16
Fix/test command

anulman:
@bendemboski here are the first steps to fix the borked test command. Having an issue with `tagName`:

image


anulman:
Closing; fixed in #188

PR #193, opened 2017-03-16
Fix module load paths

### LINKED ISSUE #175: Uncaught Error: Cannot find module 'devtron'

### LINKED ISSUE #192: `electron-protocol-serve` should not be pinned to v1.1.0
[Locking `electron-protocol-serve` to v1.1.0](https://github.com/felixrieseberg/ember-electron/blob/master/blueprints/ember-electron/index.js#L68) prevents users from getting improved/compatible new versions. I assume this was a mistake.

bendemboski:
This fixes #175 (which was actually an inability to require *any* dependencies). The problem was that we switched over to using a custom 'serve://' protocol to load index.html via electron-protocol-serve, but when loading index.html via a non-file protocol, electron's renderer/init.js doesn't know where it is in the filesystem, so it can't set up the node module search paths correctly.

So, this change adds some code to our shim-header.js to check to see if we were loaded by a non-file protocol, and if electron-protocol-serve set the ELECTRON_PROTOCOL_SERVE_INDEX environment variable giving us the path to index.html, and if so, runs some copypasta'ed code from electron's renderer/index.js to set up the module search paths correctly.

This also fixes #192

### DIFF
--- a/lib/resources/shim-head.js
+++ b/lib/resources/shim-head.js
@@ -2,6 +2,28 @@
 ((win) => {
   win.ELECTRON = true;
 
+  // On linux the renderer process doesn't inherit the main process'
+  // environment, so we need to fall back to using the remote module.
+  var serveIndex = process.env.ELECTRON_PROTOCOL_SERVE_INDEX || require('electron').remote.process.env.ELECTRON_PROTOCOL_SERVE_INDEX;
+  if (serveIndex && window.location.protocol !== 'file:') {
+    // Using electron-protocol-serve to load index.html via a 'serve:' URL
+    // prevents electron's renderer/init.js from setting the module search
+    // paths correctly. So this is basically a copy of that code, but using an
+    // environment variable set by electron-protocol-serve containing the
+    // filesystem path to index.html instead of window.location.
+    var path = require('path');
+    var Module = require('module');
+
+    global.__filename = path.normalize(serveIndex);
+    global.__dirname = path.dirname(serveIndex);
+
+    // Set module's filename so relative require can work as expected.
+    module.filename = global.__filename;
+
+    // Also search for module under the html file.
+    module.paths = module.paths.concat(Module._nodeModulePaths(global.__dirname));
+  }
+
   // Store electrons node environment injections for later usage
   win.moduleNode = win.module;
   win.processNode = win.process;
--- a/package.json
+++ b/package.json
@@ -86,7 +86,7 @@
     "broccoli-string-replace": "^0.1.1",
     "chalk": "^1.1.0",
     "electron-forge": "^2.8.0",
-    "electron-protocol-serve": "^1.1.0",
+    "electron-protocol-serve": "^1.3.0",
     "ember-cli-babel": "^5.1.7",
     "ember-cli-version-checker": "^1.1.6",
     "ember-inspector": "2.0.4",
--- a/blueprints/ember-electron/index.js
+++ b/blueprints/ember-electron/index.js
@@ -65,7 +65,7 @@ class EmberElectronBlueprint extends Blueprint {
       // Production dependencies
       save: true,
       verbose: true,
-      packages: ['electron-protocol-serve@1.1.0', ...deps]
+      packages: ['electron-protocol-serve@^1.3.0', ...deps]
     }));
   }
 
--- a/ember-electron/lib/helper.js
+++ b/ember-electron/lib/helper.js
@@ -1,3 +1,3 @@
 module.exports = function() {
-  return;
+  return 'helper';
 };
--- a/tests/unit/end-to-end-test.js
+++ b/tests/unit/end-to-end-test.js
@@ -14,3 +14,9 @@ test('node modules are accessible', function(assert) {
   let { fill } = requireNode('lodash/array');
   assert.deepEqual(fill([0, 0, 0], 1), [1, 1, 1]);
 });
+
+test('local modules are accessible', function(assert) {
+  // ../../ because we are running out of ember/tests/index.html
+  let helper = requireNode('../../lib/helper.js');
+  assert.equal(helper(), 'helper');
+});


package.json
```diff
@@ -86,7 +86,7 @@
     "broccoli-string-replace": "^0.1.1",
     "chalk": "^1.1.0",
     "electron-forge": "^2.8.0",
-    "electron-protocol-serve": "^1.1.0",
+    "electron-protocol-serve": "^1.2.0-2",
```
jacobq: This should work, but I think this was [republished as 1.3.0](https://github.com/pichfl/electron-protocol-serve/releases) during troubleshooting, so we may as well bump this to match.

bendemboski: Oh, you're right. I'll fix it and then merge.

index.js
```diff
@@ -67,9 +67,23 @@ module.exports = {
...
 
+      // Write our <base> tag to index.html so electron-protocol-serve will
+      // serve our resources correctly
+      tree = replace(tree, {
+        files: ['index.html', 'tests/index.html'],
+        pattern: {
+          match: /^(\s*)<head>/m,
+          replacement: [
+            '$1<head>',
+            `$1  <base href="serve://dist">`,
```
jacobq: Just curious, why are you using a string template (`) here instead of a literal (')?

package.json
bendemboski: copypasta from somewhere else where I was filling in a template value

bendemboski: Er, I mean, I don't know what you're talking about, there are no backticks there (yay rebasing 😉 )

index.js
bendemboski: copypasta from somewhere else where I was filling in a template value

bendemboski: Er, I mean, I don't know what you're talking about, there are no backticks there (yay rebasing 😉 )

lib/resources/shim-head.js
```diff
@@ -2,6 +2,25 @@
 ((win) => {
   win.ELECTRON = true;
 
+  if (process.env.ELECTRON_PROTOCOL_SERVE_INDEX && window.location.protocol !== 'file:') {
```
jacobq: Wouldn't it be better to use something like `=== 'serve:'` here? What if the window navigates to an `http://` URL, for example? I don't think we'd want to add this back in then since I thought the original, upstream code only applied to `file://`.

bendemboski: I don't want to check for `serve:` because `electron-protocol-serve` allows the user to customize the protocol they use. `serve:` is just the default. I suppose I could jam that into yet another environment variable, but I don't think it's necessary.

This shim is the very first piece of script that runs on the page (unless the user does something really inadvisable), and only runs once, so at this point we can be confident that `window.location` is the URL of `index.html`. So there's no risk of it having changed so some HTTP URL or something.

bendemboski: I suppose somebody could be using electron to load `index.html` over the network, but then they are pretty far outside `ember-electron`'s intended use case, so I'd like to understand what they are doing and why before writing code to try to anticipate and guess at it.

bendemboski: Note that this also allows the use case that the user has *two* ember apps bundled up in their electron app, and is using two different custom protocol handlers to serve them.

bendemboski: Something something ember engines something something ;)

jacobq: I suppose it does include a check for `process.env.ELECTRON_PROTOCOL_SERVE_INDEX` so this should only apply when using the custom protocol(s). Would it make it to messy to check against a list of well-known protocols?
e.g. replace `window.location.protocol !== 'file:'` with `isCustomProtocol(window.location.protocol)`
```
const nonCustomProtocols = ['http:', 'https:', 'ftp:', 'file:']; 
function isCustomProtocol(protocol) {
    return nonCustomProtocols.indexOf(protocol) === -1;
}
```

pichfl: I should not have made it possible to customize that protocol. However, in its current iteration the function return the name of the protocol, if that is any help. so calling `protocolServe()` returns `serve`.

bendemboski: @jacobq I get what you're saying, and perhaps further iterations down the road to make this more robust would make sense. But consider:

1. This logic only run's in Ember's `index.html` (because that's the only place we inject it)
1. It only has any effect if the renderer process accesses node modules
1. There's really no scenario where loading `index.html` from something other than a file protocol (whether literally `file:` or a custom protocol) *and* using node modules makes any kind of sense, or is not a gross violation of good security practices

So for those reasons I'm inclined to leave it as-is now, and if we run into any problems, we can iterate on it later. Since this file is in our addon, and not in the blueprint, making changes to it is easy and doesn't require the host app to modify any of their files.

jacobq:
Testing from scratch, I still get:
```
ember.debug.js:6001 DEBUG: -------------------------------
ember.debug.js:6001 DEBUG: Ember      : 2.12.0
ember.debug.js:6001 DEBUG: Ember Data : 2.12.0
ember.debug.js:6001 DEBUG: jQuery     : 3.2.0
ember.debug.js:6001 DEBUG: -------------------------------
module.js:472 Uncaught Error: Cannot find module 'devtron'
    at Module._resolveFilename (module.js:470)
    at Function.Module._resolveFilename (/home/jquant/temp/junk/node_modules/electron-prebuilt-compile/node_modules/electron/dist/resources/…:35)
    at Function.Module._load (module.js:418)
    at Module.require (module.js:498)
    at require (internal/module.js:20)
    at installDevtron (reload.js:41)
    at HTMLDocument.document.addEventListener (reload.js:96)
fs.js:1318 Uncaught Error: watch /home/jquant/temp/junk/node_modules/electron-prebuilt-compile/node_modules/electron/dist/resources/electron.asar/renderer ENOTDIR
    at exports._errnoException (util.js:1022)
    at FSWatcher.start (fs.js:1316)
    at Object.fs.watch (fs.js:1341)
    at watch (reload.js:33)
    at fs.stat (reload.js:81)
    at ELECTRON_ASAR.js:272
    at _combinedTickCallback (internal/process/next_tick.js:67)
    at process._tickCallback (internal/process/next_tick.js:98)
```

Steps to reproduce on my linux x64, node 6.10.0 system:
```
git clone git@github.com:isleofcode/electron-forge.git
cd electron-forge
git checkout feat/make-platforms
npm install
npm link
cd ..

git clone git@github.com:felixrieseberg/ember-electron
cd ember-electron
git checkout fix-require-path
npm link electron-forge
npm install
npm link
cd ..

ember new junk
cd junk
ember install file:///path/to/ember-electron
npm link ember-electron
npm install
ember g ember-electron
ember electron
```

jacobq:
Summarizing my findings (/ slack chatter):

* Module path problem appears to still be present (on linux) after these changes 
* We set up `electron-protocol-serve` from `ember-electron/lib/index.js` which executes in the main process, not the render process
* We check `process.env.ELECTRON_PROTOCOL_SERVE_INDEX` in the head shim, which is inside the renderer process
* Some OSes (notably linux) doesn't share the env here, so this var is `undefined`
* This can be avoided by checking `electron.remote.process.env` for those cases (i.e. when the other appears unset)

PR #194, opened 2017-03-16
fix(blueprint): Unpin devtron & electron-protocol versions

jacobq:
Although I think this is only a problem when running the blueprint (e.g. not if re-running `npm install` later) it still looks like a mistake that we ought to correct.

BTW, is there a reason we don't list `devtron` in `package.json`'s `dependencies`? It is required to use all of the features of `ember-electron`. I know we add it with the blueprint, but if someone, for example (and this may be a bogus use case), was inspecting our `package.json` to determine our dependencies such as to check for modules using restrictive licenses, then they could get mislead.

### DIFF
--- a/blueprints/ember-electron/index.js
+++ b/blueprints/ember-electron/index.js
@@ -40,12 +40,12 @@ module.exports = class EmberElectronBlueprint extends Blueprint {
       .then(() => npmInstall.run({
         saveDev: true,
         verbose: false,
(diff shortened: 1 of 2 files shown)


jacobq:
Looks like Ben also addressed this in #193 so I'm going to close this one in favor of his.

bendemboski:
Oh, I didn't unpin `devtron`, but maybe @anulman can in his branch so we don't conflict it yet again!

PR #205, opened 2017-04-03
Fix live reload path (don't append extra `ember`)

jacobq:
This addresses an error (see in Electron/Chrome dev tools console) that occurs on start-up and looks like:
```
Uncaught Error: watch /my-app/tmp/electron_livereload-O7Q8rQnz.tmp/ember/ember ENOENT
```

See also #182

### DIFF
--- a/app/electron/reload.js
+++ b/app/electron/reload.js
@@ -15,10 +15,8 @@
    *
    * @param sub directory
    */
-  let watch = function(...sub) {
-    let dirname = path.join(__dirname, ...sub);
-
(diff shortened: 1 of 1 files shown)


felixrieseberg:
Woops, I'm super sorry - I didn't see your pull request. I'm so very sorry. I fixed this error together with a few other ones over in #207, would you mind taking a look?

jacobq:
No worries; looks like you addressed the same problem in your fix (different but functionally equivalent). That was all I wanted :)

PR #217, opened 2017-04-08
README.md update `:package` vs `:make`

jacobq:


felixrieseberg:
That's technically not correct - `package` creates binaries, `make` creates installers.

jacobq:
Ah, I suppose that is true. My main concern was that it wasn't listing the `:make` command, which I think is rather important. Feel free to edit to your liking, otherwise maybe I'll take another shot at it later (swamped right now).

felixrieseberg:
Thanks, changed!

Issue #224, opened 2017-04-11
Ember Inspector installed in temporary location


jacobq:
For some reason, it seems that an old temp path (on generated on last `ember electron` run) is used to load the Ember Inspector. I think what is happening is that after we install it (via `addDevToolsExtension`) electron is remembering the path we supplied, which is a temporary one and changes from one execution to another.

Indeed, looking at `~/.config/Electron/DevTools\ Extensions` (path may vary by platform -- see `app.getPath('userData')`) shows that `devtron` is located in the `node_modules` folder at the root of an ember-electron project, but `ember-inspector` is in a `node_modules` folder inside of a `tmp/electron_livereload-XXX.tmp` folder.

Neither of these is truly "right" since this folder is shared by multiple projects, but I think we should at least attempt to use a non-temp location to eliminate the error message. If possible, it'd be nice to prefer a globally installed version too, if one is available.

## Steps to reproduce

* Platform: Linux x64
* `node` v6.10.2
* `npm` v3.10.10
* `ember-electron` v2.0.1
* `electron-prebuilt-compile` v1.6.2

```
ember new junk
cd junk
ember install ember-electron
npm install
bower install
ember electron
```

Running the below script in the devTools console uninstalls the Ember Inspector, which appears to temporarily alleviates the problem (doesn't show up on next build, but comes back after that):
```
var bw = requireNode('electron').remote.BrowserWindow;
var ei = bw.getDevToolsExtensions()['Ember Inspector'];
if (ei) {
    bw.removeDevToolsExtension('Ember Inspector');
}
```

## Sample output:

```
$ ember electron

"ember-welcome-page" was detected in your devDependencies!

Please note that this addon only works in development environment
and will not render in production mode. It is safe to uninstall
this addon once you removed the {{welcome-page}} template tag.


ember-electron/.eslintrc.js: line 1, col 1, 'module' is not defined.

1 error

===== 1 JSHint Error

New output path: /path/to/junk/tmp/electron_livereload-GvsxmhuQ.tmp

Build successful - 5156ms.



Slowest Nodes (totalTime => 5% )              | Total (avg)         
----------------------------------------------+---------------------
Babel (16)                                    | 4832ms (302 ms)     

Starting Electron...
Reading /last/run/project/tmp/electron_livereload-hZQ82TnP.tmp/node_modules/ember-inspector/dist/chrome/manifest.json failed.
[...]

PR #227, opened 2017-04-12
Revamp build pipeline

bendemboski:
This is a significant re-write of the build pipeline, but it does not break compatibility. The blueprint is unaffected, and command functionality has been added, but not changed/removed. The only functional change (other than adding commands and command options) is that `ember electron` will build into `electron-out/project` and then run from there, rather than a temp directory, and leaves the output there when it exits. This matches `ember serve`'s behavior, and is nice for inspecting the files after running (especially if the app is crashing on startup!).

Code-architecturally, I moved us more in the direction of ember-cli by putting most of the work into tasks (and a couple of models), and making the commands mostly just a wrapper for the tasks (except I didn't really touch `ember electron` and `ember electron:test`). I also wrote a whole crap-ton of unit tests.

I added a build pipeline guide that describes how it works from a user/cli standpoint now, so I won't duplicate that here.


jacobq:
Might this also fix https://github.com/felixrieseberg/ember-electron/issues/224?

bendemboski:
Yeah, I guess it might since we aren't using a temp dir anymore when running `ember electron`.

felixrieseberg:
...what a beast of a PR. Hats off to you, my friend, I'm pretty impressed.

PR #239, opened 2017-04-20
Fix package json

bendemboski:
Fixes a few bugs and adds some checks and end-to-end tests so hopefully we'll see fewer blueprint-related regressions.


pichfl:
If the `.eslintrc.js` files in tests and the blueprint do not work and need to be replaced by the inline comments we can safely remove those files.

felixrieseberg:
Tons of good stuff. Should we just merge this or...?

bendemboski:
Not just yet -- I'm right on the verge of getting it to build on Appveyor...

bendemboski:
Okay @felixrieseberg it looks like this is all good to go. I don't think I have permission to make Appveyor builds run automatically as one of the checks -- can you set that up?

felixrieseberg:
Will set it up!

PR #248, opened 2017-05-01
Update to latest electron-forge

bendemboski:
3.0.0 with cross-platform make targets and various other improvements!

### DIFF
--- a/lib/commands/test.js
+++ b/lib/commands/test.js
@@ -20,7 +20,11 @@ module.exports = TestCommand.extend({
 
   rmTmp() {
     if (process.platform === 'win32') {
-      console.log(require('child_process').execSync('handle').toString());
+      // There is some kind of race condition where we intermittently get EBUSY
+      // deleting the temp directory on windows, which causes the test to fail.
+      // Unfortunately adding logging to try to track it down makes it go away,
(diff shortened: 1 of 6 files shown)


lib/commands/test.js
```diff
@@ -20,7 +20,11 @@ module.exports = TestCommand.extend({
 
   rmTmp() {
     if (process.platform === 'win32') {
-      console.log(require('child_process').execSync('handle').toString());
+      // There is some kind of race condition where we intermittently get EBUSY
+      // deleting the temp directory on windows, which causes the test to fail.
+      // Unfortunately adding logging to try to track it down makes it go away,
+      // so let's just delay for a bit as a dirty workaround.
+      require('child_process').execSync('ping 127.0.0.1 -n 2 > nul');
```
malept: Uh, does this really work consistently? 😵 If so, we may have to borrow this for Forge tests.

bendemboski: So far. Before I added it, about 1 out of every 2 tests would fail, and with this I've run 5 or 6 with no failures. I wanted to use `timeout 2` but it barfs if it's not run with a tty as stdin and I was having trouble convincing any of the child process `*sync` commands to forward stdin for some reason.

But yeah, this appears to work consistently.

PR #266, opened 2017-06-02
Add a FAQ section on same-origin considerations

bendemboski:


pichfl:
While this is one way of solving this, I'd also suggest that a user should consider adding the appropriate headers to their API responses if possible.

bendemboski:
@pichfl that's a good idea, I'll open another PR with that included...

PR #274, opened 2017-06-23
Fix node_modules symlink again

bendemboski:
Fix a couple more cases where we can delete content out of the root project's node_modules:

1. Build with symlinking turned on (i.e. run ember electron) and then build with it turned off (e.g. run ember build)
2. Build with symlinking turned on (i.e. run ember electron) and then assemble (i.e. ember electron:assemble)

### DIFF
--- a/lib/models/assembler.js
+++ b/lib/models/assembler.js
@@ -10,6 +10,7 @@ const CoreObject = require('core-object');
 const SilentError = require('silent-error');
 const chalk = require('chalk');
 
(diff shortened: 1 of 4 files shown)


bendemboski:
The appveyor build is failing because 5b8e01f5a52edaf17cab931d27c1a531651400ce, but should pass once #273 is merged and I can rebase this onto it.

felixrieseberg:
Sorry, I was on vacation. Let's merge all of this 😅

PR #284, opened 2017-07-17
Move test main.js to ember-electron/test-main.js

ursm:
Files in the `tests` directory are transpiled by Babel. Depending on the application's Babel setting, `require()` in test `main.js` may be truncated. Orginally this file does not need to be combined into `assets/tests.js`, so put it outside the `tests` directory.

Closes #258

PR #287, opened 2017-08-02
Exclude test-main.js from assemble output

ursm:
Fixed that test-main.js was unintentionally included in the output of the assemble tree with the modification of #284.


felixrieseberg:
:rocket:

PR #289, opened 2017-08-10
Improved yarn support

stopdropandrew:
**New project setup:**
`ember new ee-test --yarn`
`cd ee-test/`
`ember install ember-electron`
`ember electron:package`
```
log.gauge.isEnabled is not a function
TypeError: log.gauge.isEnabled is not a function
    at Object.<anonymous> (/.../ee-test/node_modules/npm/node_modules/npmlog/log.js:57:33)
...
```

**Initial workaround:**
`rm -rf node_modules`
`yarn upgrade`

**But I want electron-packager to use yarn!**
`sed -i '' 's/\"electronPackagerConfig\": {}/\"electronPackagerConfig\": { \"packageManager\": \"yarn\" }/' ember-electron/electron-forge-config.js`
`ember electron:package`
```
...
⠧ Packaging ApplicationCommand failed: yarn install --production --no-bin-links
warning electron-prebuilt-compile > electron-compilers > jade@1.11.0: Jade has been renamed to pug, please install the latest version of pug instead of jade
warning ember-electron > ember-inspector > ember-new-computed > ember-cli-babel > broccoli-babel-transpiler > babel-core > minimatch@2.0.10: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
warning ember-electron > electron-forge > fs-promise@2.0.3: Use mz or fs-extra^3.0 with Promise Support
warning electron-prebuilt-compile > electron-compilers > jade > transformers@2.1.0: Deprecated, use jstransformer
warning ember-electron > electron-forge > zip-folder > archiver > glob > minimatch@0.3.0: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
warning ember-electron > npmi > npm > request > node-uuid@1.4.8: Use uuid module instead
warning ember-electron > electron-forge > electron-windows-store > flatten-packages > wrench@1.5.9: wrench.js is deprecated! You should check out fs-extra (https://github.com/jprichardson/node-fs-extra) for any operations you were using wrench for. Thanks for all the usage over the years.
warning ember-electron > electron-forge > nugget > progress-stream > through2 > xtend > object-keys@0.4.0: 
error /Users/andrew/Library/Caches/Yarn/v1/npm-electron-1.7.5-065a3102bf8b87102df50c50985fefe6c569045b: Command failed.
Exit code: 1
Command: sh
Arguments: -c node install.js
Directory: /Users/andrew/Library/Caches/Yarn/v1/npm-electron-1.7.5-065a3102bf8b87102df50c50985fefe6c569045b
Output:
module.js:442
    throw err;
    ^

Error: Cannot find module 'extract-zip'
    at Function.Module._resolveFilename (module.js:440:15)
    at Function.Module._load (module.js:388:25)
    at Module.require (module.js:468:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/Users/andrew/Library/Caches/Yarn/v1/npm-electron-1.7.5-065a3102bf8b87102df50c50985fefe6c569045b/install.js:9:15)
    at Module._compile (module.js:541:32)
    at Object.Module._extensions..js (module.js:550:10)
    at Module.load (module.js:458:32)
    at tryModuleLoad (module.js:417:12)
    at Function.Module._load (module.js:409:3)

Error: Command failed: yarn install --production --no-bin-links
...
```

There are a handful of intertwined issues here.
* EE ignores shrinkwrap and yarn lock files when assembling (causes the issue in electron-packager above)
* EE always uses npm to install production dependencies when assembling
* checking in yarn.lock means that tests are not run against the same dependencies that users are running against


lib/tasks/assemble.js
```diff
@@ -1,7 +1,8 @@
 'use strict';
 
 const chalk = require('chalk');
-const install = require('rsvp').denodeify(require('npmi'));
+const pruner = require('rsvp').denodeify(require('electron-packager/prune').pruneModules);
```
bendemboski: I don't think this is really a public API, so I'm kinda uncomfortable using it like this. My preference would be to copy the code that we need into this project, but failing that, at the very least list `electron-packager` as one of our dependencies, rather than just implicitly relying on `electron-forge` to make it available to us.

stopdropandrew: Very fair, I felt uncomfortable doing it as well. Wanted to show intent and happy to make any changes.

lib/tasks/assemble.js
```diff
@@ -1,7 +1,8 @@
...
+const { hasYarn } = require('electron-forge/dist/util/yarn-or-npm');
```
bendemboski: I also think maybe here we should just use the `yarn-or-npm` package directly, rather than using a private API in `electron-forge`. AFAIKT `electron-forge` only wraps it so it can introduce an environment variable for its automated testing. We don't currently need that, and if we did, I'd rather have our own implementation so we don't break if `electron-forge` changes how they do it.

stopdropandrew:
This should be good to go.

felixrieseberg:
Hey! I have one last comment: I want this to work with Yarn as much as possible, but I don't want yarn to be the _default_ runner.

Electron is cool (and better than a PWA) because you can use it with native code. Native modules and Yarn still aren't friends (https://github.com/yarnpkg/yarn/issues/756), so until that is the case, I'd like to stick with npm.

As it is right now, the `package.json` scripts require yarn, which I'm not suuuuper excited about. Could we go with `npm` or at least `yarn-or-npm`?

BenV:
@felixrieseberg Is this what you had in mind?

felixrieseberg:
Great, bank you!

PR #293, opened 2017-08-12
Fix live-reload

synaptiko:
I installed fresh version of ember-electron and live-reload didn't work. I found out that `window.process` is undefined and `window.processNode` should be used instead.

Maybe there can be more bugs like this?

### DIFF
--- a/app/electron/reload.js
+++ b/app/electron/reload.js
@@ -1,5 +1,5 @@
 function setupLivereload() {
-  const process = window ? window.process : null;
+  const process = window ? window.processNode : null;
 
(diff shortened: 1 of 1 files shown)


synaptiko:
Failed build looks to be related to this transitive dependency: `macos-alias`.

bendemboski:
The problem is that for some reason we are running with a too-old version of yarn on travis. I haven't had a chance to investigate why, but we need to be running with a much newer version than `v0.17.8` -- ideally whatever the latest is.

felixrieseberg:
This PR is great, thank you.

PR #294, opened 2017-08-13
Rewrite index-electron.html before reloading tests

BenV:
After upgrading to EE2 we noticed that the test server would reload to a blank page after modifying a file due to the `index-electron.html` file getting wiped out by the build. This also happens with a  brand new Ember Electron 2 project. 

This (admittedly cheesy) workaround that has saved us some pain, and I figured you might be interested in it.

I'm sure that you all can come up with a more elegant solution, but hopefully this at least demonstrates what the issue is. Please let me know if you need/want and more information. Thanks!


bendemboski:
Thanks a lot for this! Fixing this issue has been in the back of my mind for a while, but I haven't had time to look into it. (Re-)generating the file on-demand is a pretty good idea, although now that I got me thinking about it, I think there might be a cleaner way.

The fundamental problem is that `index-electron.html` needs information that only testem can provide, so I don't think there's a good way of generating it as part of the broccoli pipeline. That means broccoli doesn't know about it, so anytime it performs a sync it thinks it's a file that's been deleted from the source tree, and deletes it from the output directory.

All the files are being served through [electron-protocol-serve](https://github.com/pichfl/electron-protocol-serve/), which can be configured with both a working directory _and_ an index path. So, I _think_ it would work to modify `test-main.js` to copy `index-electron.html` to a temp directory outside the broccoli-managed directory and then modify [this](https://github.com/felixrieseberg/ember-electron/blob/master/blueprints/ember-electron/files/ember-electron/test-main.js#L31) code to point `electron-protocol-serve` there. Broccoli would still delete the original, but by then we wouldn't care.

Interested in giving this solution a shot?

BenV:
Sounds like a good solution to me, I'll work on that today and get back to you, thanks!

BenV:
@bendemboski - Let me know if this approach is close to what you were thinking, it seems to work fine. I'm not sure if it is overkill to clean up the temporary file on exit or not, but it seemed appropriate to at least remove it upon a clean exit. I also wasn't sure whether to squash these commits or not, I can do that along with any other cleanup if you would like.

BenV:
Will investigate test failure

BenV:
I fixed the test, but it seems odd that it previously had to use `../../` at all since the cwd should be set to the ember app directory, and that cwd should not have changed. Does this seem OK? I'd hate to subtly change behavior and then cause `requireNode` to fail for people.

bendemboski:
The problem is that for some reason we are running with a too-old version of yarn on travis. I haven't had a chance to investigate why, but we need to be running with a much newer version than v0.17.8 -- ideally whatever the latest is.

bendemboski:
The problem causing the test failure is that for some reason we are running with a too-old version of yarn on travis. I haven't had a chance to investigate why, but we need to be running with a much newer version than v0.17.8 -- ideally whatever the latest is.

ember-electron/test-main.js
```diff
@@ -14,23 +15,29 @@ let [, , indexUrl] = process.argv;
...
-  indexPath = indexPath.slice(1);
+  originalIndexPath = originalIndexPath.slice(1);
 }
-indexPath = resolve(indexPath);
+
+// Copy index-electron.html to a location not managed by broccoli so that it
+// doesn't get deleted when a change is detected.
+originalIndexPath = resolve(originalIndexPath);
+let emberAppDir = resolve(dirname(originalIndexPath), '..');
+let indexPath = resolve(emberAppDir, '../..', basename(originalIndexPath));
```
bendemboski: I think the best way would be to just create a new temp directory using the `tmp` module. The `tmp` module automatically deletes its directories when the process exits.

BenV: Ah, perfect. I'll go ahead and do that, thanks!

bendemboski:
> I fixed the test, but it seems odd that it previously had to use ../../ at all since the cwd should be set to the ember app directory, and that cwd should not have changed. Does this seem OK? I'd hate to subtly change behavior and then cause requireNode to fail for people.

Shoot, you're right, I had forgotten that we [set up the module search paths](https://github.com/felixrieseberg/ember-electron/blob/master/lib/resources/shim-head.js#L7) we do it based on the location of `index.html`. It's a bit of a dirty workaround, but maybe for now we could try rewriting that environment variable in `test-main.js` to point to the original path of `index-electron.html`?

BenV:
OK, I think this should be working. Let me know what you think, and if you want me to squash the commits since I've been scrambling a bit 😄

ember-electron/test-main.js
```diff
@@ -14,23 +16,30 @@ let [, , indexUrl] = process.argv;
...
-  indexPath = indexPath.slice(1);
+  originalIndexPath = originalIndexPath.slice(1);
 }
-indexPath = resolve(indexPath);
+
+// Copy index-electron.html to a location not managed by broccoli so that it
+// doesn't get deleted when a change is detected.
+originalIndexPath = resolve(originalIndexPath);
+let emberAppDir = resolve(dirname(originalIndexPath), '..');
+let tmpDir = tmp.dirSync({ dir: resolve(emberAppDir, '../../'), unsafeCleanup: true }).name;
```
bendemboski: Do you see a benefit to specifying the `dir` option here? It seems to me it would work just as well not to, so the temp directory is created off in the system default tmp location, and then we don't have to bake an assumption about how the parent directories are set up into this file.

BenV: My concern was just with permissions writing to the system tmp directory, but my guess is if you can't write to that you have bigger problems anyway, so I'll go ahead and remove it.

BenV:
As per your comment on the other PR, I guess I should update the blueprint as well?

ember-electron/test-main.js
bendemboski: Hmm...yeah, I guess the only time we use the system tmp location is when running the tests in this addon (i.e. not in a client app). But I think you're right that not having access to `tmp` is probably going to be fatal for other `electron`-related reasons...in any case we'll have to move the `tmp` module from `devDependencies` to `dependencies`.

bendemboski:
Yeah, I just files an issue with some ideas for how to remove the duplication. If you're feeling super contributory, you could take a look at that too 😄 Otherwise, I can probably find time before too long...

felixrieseberg:
@BenV @bendemboski This seems to work and I'd be happy to merge it in - it works, and it didn't before! Are you cool with me hitting that merge button?

BenV:
@felixrieseberg Hold up, need to update the blueprint first 😄 , fool me once...

bendemboski:
BTW, the more little gotchas we run into with this fix, the more I think we need to find a way to just get the `index-electron.html` into the broccoli pipeline and get rid of [this code](https://github.com/felixrieseberg/ember-electron/blob/master/test-runner.js#L51) code)...I _think_ it would probably be doable using `broccoli-string-replace` to fix the testem path [here](https://github.com/felixrieseberg/ember-electron/blob/master/lib/utils/assemble-tree.js#L64) and the `contentFor` hook to inject the `getTestemId` script [here](https://github.com/felixrieseberg/ember-electron/blob/master/index.js#L62), both conditional on `process.env.EMBER_ENV === 'test'`. Then broccoli would be generating the `index.html` file, and all these acrobatics we're trying to do would be unnecessary.

If you just want to push this PR through, I'm totally fine with that, and I can just file an issue describing ☝️ to get to when we can. If you are interested in exploring this better-and-hopefully-as-simple-as-I-described-(-yeah-right-)-solution, I'm happy to advise/troubleshoot as needed. Here, or you can find me on the Ember community slack.

Either way is cool with me -- just let me know.

BenV:
Yeah, it is funny that as we ran into more and more things my original super hacky solution started looking better and better to me 😄  - I will try to find some time in the near future to investigate a pure broccoli solution in the near future, as I think that would be a good learning experience. Thanks for all the assistance on this, hopefully this latest iteration is acceptable.

ember-electron/test-main.js
bendemboski: Looks like you forgot to remove the `dir` option

bendemboski: Oh, also  `unsafeCleanup` isn't necessary since we aren't creating any sub-folders

BenV: Oh sorry, from your comment I thought you wanted me to keep the `dir` option. I'll remove that and `unsafeCleanup` - I wasn't sure if that was needed since we're creating a file in the tmp dir. Good to know.

bendemboski: Oh yeah, I can see how that wasn't clear -- my bad!

bendemboski:
Hmm...not sure why your appveyor build failed -- might be a transitory failure? Maybe try cherry-picking 0182501f793788330d629e1535fc613ccf839da7 (or just manually editing to match), which should fix your travis build, and then hopefully the appveyor build will pass and the PR will be green?

bendemboski:
Sorry I keep forgetting to respond to your question about squashing! If the build passes, go ahead and squash.

BenV:
OK, cherry-picked and squashed, and builds are green!

bendemboski:
🎉  thanks for the contribution!

PR #296, opened 2017-08-14
Fix the backgroundThrottling option for tests

BenV:
The `backgroundThrottling` option needs to be part of `webPreferences`: https://github.com/electron/electron/blob/master/docs/api/browser-window.md#new-browserwindowoptions

### DIFF
--- a/ember-electron/test-main.js
+++ b/ember-electron/test-main.js
@@ -46,7 +46,9 @@ app.on('ready', function onReady() {
   mainWindow = new BrowserWindow({
     width: 800,
     height: 600,
-    backgroundThrottling: false,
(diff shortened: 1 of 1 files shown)


felixrieseberg:
Oh, good call, thanks!

bendemboski:
This will only affect running the tests for this project, it won't affect anything in apps using this addon. The same change also needs to me made to the copy of this file in the blueprint. I'd like to fix it so we don't have two very similar copies of this file sometime, but that's how things are now for the time being...

BenV:
Ah, my bad, I didn't realize there were two copies. I'll open a pull for the blueprint as well.

PR #330, opened 2017-12-13
Do a minor overhaul of our electron:test setup

### LINKED ISSUE #307: ember electron:test --serve hangs on reload
On a fresh install of Ember with ember-electron,  `ember electron:test --serve` loads the electron renderer and runs the tests, but when a file is changed, the renderer attempts to refresh and instead hangs on a blank screen, with the tests never indicating they are rerun.

Steps:

```
ember new test-electron
cd test-electron
ember install ember-electron
ember electron:test --serve
```

Then trigger a restart by saving a file. Test browser will hang on blank screen. 

Ember-cli: 2.14.2
Ember-electron: 2.3.1
macOS Sierra 10.12.6

Please let me know if any more information is needed or if I can take any steps to help identify the source of this issue.

bendemboski:
The goal here is to get rid of all the file hacking in test-runner.js, and have the brocolli pipeline manage all of our file operations. This involved moving the getTestemId() function that allow the testem client to communicate back to the server into a shim, and then doing some trickery in the main process to redirect the testem.js request to the testem server so we don't have to rewrite the script tag, which we can't do from broccoli because we don't know the testem URL yet.
    
In doing this, I realized that the tight coupling between test-runner.js and test-main.js is problematic because test-runner.js is used directly from the ember-electron package, while test-main.js is generated by the blueprint in the app project. So upgrades are hairy. So, this moves the test-main.js code out of the blueprint and into ember-electron so we can update it without requiring users to re-run the blueprint. If users want to customize test-main.js, they still can, but then they will be responsible for maintaining their project across updates to test-runner.js.
    
I also left the original test-runner.js in place (with a minor change to accomodate the assemble-tree.js change) so projects that use it, and therefore also have an old-style test-main.js, won't break when updating to this version of the addon.

Fixes #307


bendemboski:
@felixrieseberg and @anulman I'd like your thoughts on the upgrade path here. As it stands, if people just update to a version of ember-electron with this change, their test commands will continue to function because I left the old `test-runner.js` in place, but of course they won't get the bugfix. I they re-run the blueprint to regenerate `testem-electron.js`, `test-main.js` and the new `test-runner.js`, they'll be all up-to-date.

So, I think our options are:

1. Leave it exactly as-is and implement an end-to-end test that verifies that the addon can run the test command in an app with a pre-this-change blueprint.
2. Leave it as-is, implement end-to-end test, but add some code to the old `test-runner.js` that prints a loud deprecation warning with instructions for updating.
3. Delete the old `test-runner.js` and release this as a major version bump with release notes indicating that a re-run of the blueprint is required.
4. Think about how to implement some upgrade tooling. I think maybe we could make the `ember electron:test` command inspect `test-main.js` and if it matches the blueprint `test-main.js` exactly, just apply the blueprint updates on the fly (`test-main.js` and `testem-electron.js`). This would take some effort, but is probably doable.

I think I would probably vote for (2) and then plan to delete the old `test-runner.js` whenever we next decide to cut a major version.

Thoughts?

felixrieseberg:
As always, 👏 👏 👏 ! I'd personally favor option 2, just because it informs everyone involved but requires the least amount of additional work.

lib/test-support/test-runner.js
```diff
@@ -0,0 +1,63 @@
...
+  let indexObj = url.parse(fileUrl(indexPath));
+  indexObj.query = testPageQuery;
+  indexObj.query.testemId = id;
+  let testUrl = url.format(indexObj);
+  // On windows the testUrl argv is truncated before the first '&' by the time
+  // it reaches our main.js. This appears to have something to do with how
+  // electron-compile (I think) uses a batch script to invoke its cli.js, and
+  // the fact that '&' is a special shell character. So we do our own cheesy
+  // workaround.
+  testUrl = testUrl.replace(/&/g, '__amp__');
```
felixrieseberg: Holy cow, that's some excellent debugging right here 🕵️

anulman:
+1. This looks hecka-tidy

bendemboski:
Okay thanks @felixrieseberg and @anulman. I'll implement that one unit test, add the deprecation warning and merge.

PR #337, opened 2018-02-01
Update Electron Forge to 5.0.x

malept:
Since 4.1.x:

* Config for electron-rebuild (thanks to @bendemboski)
* `electron-wix-msi` support
* Snapcraft support, both making and publishing to the Snap Store

The major version bump has to do with the third party publisher API and should not affect ember-electron.


felixrieseberg:
As always: Thank you, @malept!

PR #339, opened 2018-02-26
Distinguish between "dev tools" and "inspector"

jacobq:
The term "inspector" is a little vague because of it could refer to either the browsers' "Inspect" menu command (i.e. open developer tools and show selected element) and the "Ember Inspector" (i.e. for examining details of the Ember application state). If I understand correctly, the current version of ember-electron strips devtron and the ember-inspector addons from the production build, but the dev tools are part of electron and never full removed (though can be `webPreferences: { devTools: false }`).

I'm not sure my wording changes here are the best, so feel free to edit. I just find it somewhat confusing when the term "inspector" is used to refer to the basic developer tools when it could also be referring to the Ember Inspector.


felixrieseberg:
As always: Thanks a lot 🙇

PR #342, opened 2018-03-01
Add FAQ doc for security-related questions

### LINKED ISSUE #341: Modify `app/index.html` with Content-Security-Policy
Electron v2.x emits a security warning in the console if the page does not define a `Content-Security-Policy`. Would it be possible to add something like the code below to `app/index.html` (e.g. above `{{content-for "head"}}`)? I think this would make it easier for users to discover how to address this warning without breaking their code. (Or maybe we *do* want to break potentially insecure code by default and require opt-out of protection?)

```html
    <!-- To eliminate the security warning in Electron 2.x, remove 'unsafe-eval'
         from the Content-Security-Policy specified in the <meta> below.
         Note that this may break your application if your code (or a dependency) requires these capabilities.
         In that case, you may see an error similar to this:
         Uncaught EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval'
         is not an allowed source of script in the following Content Security Policy directive...
    -->
    <meta http-equiv="Content-Security-Policy" content="script-src 'unsafe-inline' 'unsafe-eval' serve://dist/">
```

jacobq:
This is a first draft and just includes one entry,
which is regarding how to address an electron security warning
that will appear if users do not define a sufficiently strict
Content-Security-Policy.

Closes #341


jacobq:
I just realized that there are also some issues with testing when CSP is set. For example, I added the following to my `config/environment.js`:
```js
    // ...
    if (environment === 'test') {
        ENV.contentSecurityPolicy['script-src'].push("http://testemserver");
        ENV.contentSecurityPolicy['script-src'].push("http://localhost:7357");
        ENV.contentSecurityPolicy['frame-src'] = ["http://testemserver", "http://localhost:7357"];
        // ...
```

jacobq:
Perhaps better still, we can disable CSP & the security warnings during testing:
```js
ENV.contentSecurityPolicyMeta = false;
process.env.ELECTRON_DISABLE_SECURITY_WARNINGS = true;
```

bendemboski:
mmm...I don't know...running unit tests is often how I catch warnings/deprecations/etc. If somebody is doing something CSP-sketchy but only in specific routes or something, I'd hate to squash that warning in unit tests...

jacobq:
@bendemboski Good point. Then the first direction would be better. (I don't think anything's going to break because of allowing `http://testemserver` and `http://localhost:7357` in test mode but not in production.) I was just saying that if one is going to use CSP (and not be ridiculously loose with it) then one also needs to change *something* in the test env rather than leave it untouched, or the tests won't run. Maybe there is some other alternative (framing w/ certain parameters?) but none is coming to mind.

felixrieseberg:
I think the document is solid. Thank you very much for it!

PR #354, opened 2018-07-09
[WIP] Upgrade testem & ember-cli

jacobq:

PR #358, opened 2018-07-17
WIP: Upgrade deps

### LINKED ISSUE #356: Update some old dependencies
I'm sure that upgrading some of these could require code changes in `ember-electron`, and I know that chasing little details like this often turns into useless yak shaving, but it looks like a bunch of them are way way behind. I've also been having some little hiccups with `ember-inspector` that I think might be fixed by using the latest version ([and at first glance I don't see breaking changes that affect us](https://github.com/emberjs/ember-inspector/blob/master/CHANGELOG.md#v300-2018-03-08)).

```
# from npm-check-updates 
# based on master/d47826b3b4f9cad3d5797eee244339af35385e2d
# Tue Jul 17 16:07:22 CDT 2018
broccoli-file-creator       ^1.1.1  →   ^2.1.1 
broccoli-funnel             ^1.0.1  →   ^2.0.1 
broccoli-merge-trees        ^2.0.0  →   ^3.0.0 
chalk                       ^1.1.0  →   ^2.4.1 
console-ui                  ^1.0.3  →   ^2.2.2 
electron-forge              ~5.0.0  →   ~5.2.2 
ember-inspector              2.0.4  →    3.1.0 
execa                       ^0.6.3  →  ^0.10.0 
fs-extra                    ^2.0.0  →   ^7.0.0 
globby                      ^6.0.0  →   ^8.0.1 
tmp                         0.0.31  →   0.0.33 
chai                        ^3.5.0  →   ^4.1.2 
chai-as-promised            ^5.2.0  →   ^7.1.1 
electron-prebuilt-compile   1.4.15  →    2.0.4 
eslint                     ^3.16.1  →   ^5.1.0 
eslint-plugin-node          ^6.0.1  →   ^7.0.1 
mocha                       ^3.5.3  →   ^5.2.0 
mockery                     ^1.4.0  →   ^2.1.0 
qunit-dom                   ^0.6.2  →   ^0.7.1 
sinon                      ^1.17.2  →   ^6.1.3 
testdouble                  ^2.1.2  →   ^3.8.1 
```

jacobq:
Closes #356
Needs changes & testing


jacobq:
AppVeyor / Windows seems to be [getting stuck after the fast tests](https://ci.appveyor.com/project/felixrieseberg/ember-electron/build/1.0.278#L552) now :thinking:

jacobq:
I can reproduce [that](https://github.com/felixrieseberg/ember-electron/pull/358#issuecomment-408166939) locally on Windows 10 with node ^6 using `cmd.exe`
image, but when I use Git Bash (MSYS) as a terminal it doesn't get stuck here:
image (though in the screenshot here I did abort it shortly after it had moved on to the next step).

jacobq:
Hmm, switching to power-shell didn't get it unstuck, but it did mess up the unicode checkmarks ;)
Anyone have ideas for (1) which package is having trouble with the terminal or (2) what can be done to work around it?

jacobq:
Aha! I think I [found it](https://github.com/mochajs/mocha/issues/2879) -- one of the breaking changing in mocha 4.x is that it doesn't force the process to exit.

> If the mocha process is still alive after your tests seem "done", then your tests have scheduled something to happen (asynchronously) and haven't cleaned up after themselves properly. Did you leave a socket open?
> Supply the `--exit` flag to use pre-v4 behavior.

jacobq:
@felixrieseberg How confident are you in this project's test coverage? (i.e. passing tests imply that nothing is broken) If it's high then it should be OK to merge this. (TravisCI failed on node 8 but passed on 6 and 10, and it looks like it was due to a network or yarn/npm registry problem.)

.eslintrc.js
```diff
@@ -78,6 +78,23 @@ module.exports = {
...
+        node: true,
+        mocha: true
+      }
+    }, {
+      // work-around unexpected (false positive?) eslint error
+      files: [
+        'blueprints/ember-electron/index.js'
+      ],
+      rules: Object.assign({}, require('eslint-plugin-node').configs.recommended.rules, {
+        "node/no-unpublished-require": "warn"
```
jacobq: @bendemboski Do you know why this is needed? It seems too coincidental with [the `require` problem I had to work-around](https://github.com/felixrieseberg/ember-electron/pull/354#issuecomment-403650859) in https://github.com/felixrieseberg/ember-electron/pull/355 to sit well with me

bendemboski: Sure, so `no-unpublished-require` is meant to verify that the package doesn't try to use any code that isn't guaranteed to be there at runtime, e.g. accidentally adding something `require`d at runtime to `devDependencies` instead of `dependencies`. But it's 100% safe to assume that when a blueprint is running, `ember-cli` is present because `ember-cli` is what actually executes the blueprints.

This is actually very closely analogous to the use case described [here](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-unpublished-require.md#allowmodules), so I think it would be best to leave it as an error, but specify `ember-cli` under `allowModules` here.

jacobq:
@felixrieseberg I just learned about it too (from the Mocha release notes / changelog). There is also a similar tool called [`why-is-node-running`](https://github.com/mafintosh/why-is-node-running); YMMV.

.eslintrc.js
jacobq: Next question: can we easily apply this only to the `ember-cli` so we still get the benefit of the check in cases where it could actually be a problem? Should we just use a single-line comment in this one spot?

bendemboski: 👍

jacobq:
BTW, it wouldn't surprise me if the mocha hang-up issue on Windows is related to https://github.com/tjunnone/npm-check-updates/issues/136

.eslintrc.js
```diff
@@ -78,6 +78,25 @@ module.exports = {
...
+        mocha: true
+      }
+    }, {
+      // ember-cli is guaranteed to be present when running blueprints
+      files: [
+        'blueprints/**/*.js'
+      ],
+      rules: Object.assign({}, require('eslint-plugin-node').configs.recommended.rules, {
+        "node/no-unpublished-require": ["error", {
+          "allowModules": ["ember-cli"]
```
jacobq: Doesn't it seem like this could be / should be part of `'plugin:ember/recommended'`?

bendemboski: No, not in `eslint-plugin-ember` -- it's job is just to supply rules for Ember best practices. If anything it would go in `ember-cli`'s addon blueprint (which is what generates this file in addons). But I think this a pretty edgy edge case -- most addon blueprint don't subclass the `Blueprint` class (which is why ours `require` `ember-cli`), but instead do the simpler export of a hash. So I don't think it really makes sense to put extra stuff in every single addon's `eslintrc.js` to address this pretty rare edge case,

jacobq: Gotcha -- that makes me feel better about this `override`.

jacobq:
Aiight, I think I'm done working on this unless someone requests changes. One thing to note is that it doesn't include code to [modify the root `eslintrc` file like we talked about](https://github.com/felixrieseberg/ember-electron/pull/355#issuecomment-405700486) but rather keeps the old behavior of adding another `eslintrc` file in the `ember-electron` folder. I just didn't have time to get to it.

jeanduplessis:
@felixrieseberg any chance we can get these changes merged and released as a new version? We're about to embark on upgrading our Electron app and would be great to use the latest and greatest.

jacobq:
Heads up: Just rebased (to v2.9.0), upgraded again (and just as carelessly ;)), and then did `git push -f` :bomb:

felixrieseberg:
@jacobq I just made you a contributor - you've been pushing so much code that you're probably in a better position to decide when exactly you want to merge this!

The changes look good to me.

jacobq:
> I just made you a contributor

@felixrieseberg Well thanks! If only that in itself would make me a better developer ;-) In the case of this PR I would recommend landing #363 first then making a `beta` release and getting feedback before tagging as latest/stable. I don't think there are any breaking changing here but also don't trust the tests to catch every possible breakage with this many moving parts.

jacobq:
@jeanduplessis have you already done your upgrade? If I update this and ask Felix to release to a beta version would you be willing to try it out?

jeanduplessis:
@jacobq yes we did, but we're doing another release in 2 weeks so I'd be happy to give the beta a try.

jacobq:
Note: also should update default electron version (ref https://github.com/felixrieseberg/ember-electron/issues/365#issuecomment-425909309)

jeanduplessis:
@jacobq anything I can do to help move this PR along? The upgrade of electron-forge to 5.2.x is a big requirement for me as it brings in electron-packager 11.x which fixes an issue we're having with the Electron Login Helper.app not being renamed. I'm currently forcing the resolution of electron-forge to 5.2.2 but ideally would not like to do that.

jacobq:
@jeanduplessis I actually haven't looked into upgrading to the latest `electron-forge`. Could you help identify potential sticking points? (i.e. find the breaking changes and see how it affects us) If I recall, `ember-electron` uses some of `electron-forge`'s private APIs, so it wouldn't surprise me at all if a bunch of stuff breaks :/

jeanduplessis:
@jacobq using electron-forge 5.2.2 as per this PR’s proposed change is working 100%. In fact we just got our upgraded production app approved by Apple which used 5.2.2. In fact not using it fails to validate with the App Store as the ‘Electron Login Helper.app’ that is included in the packaged app doesn’t get renamed to ‘<your app name> Login Helper>.app’ causing a conflict. The newer version of electron-forge resolved that issue.

jacobq:
@jeanduplessis I finally got back to this this morning and have it ready to merge as soon as AppVeyor CI is working again. Any insights you can off to figure out why/how it's getting stuck (see https://github.com/felixrieseberg/ember-electron/pull/367) would be helpful and appreciated.

jeanduplessis:
@jacobq I’ll try and have a look a bit later this evening.

jacobq:
@jeanduplessis Not sure if you had a chance to look, but the problem seems to be caused by a compatibility problem with electron v3.x & ember-electron on Windows and the fact that `electron-forge` was using electron v3.x by default (since it is `latest`).

I think that the short-term plan is:

1. Force default use of 2.x (only affects new projects and can be changed later) to get CI passing again
2. Merge this update
3. Open a new issue to investigate and address the compatibility problem with 3.x

jacobq:
Hmm, something is going a muck again --> https://travis-ci.org/felixrieseberg/ember-electron/jobs/446329080#L819 Not sure if it's because of the dep update or something else...

jacobq:
Going to rebase/redo this and make the commit messages follow the conventional changelog standard.

PR #363, opened 2018-08-30
Fix #362 (Call _super at end of init)

### LINKED ISSUE #362: Error: The "path" argument must be of type string
Occurring with:

* ember-electron 2.9.0
* ember-cli 2.15.1
* node 9.x (also tried node 7.x)

After installing EE, I simply ran `ember electron` and got the following error:

```
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string
    at assertPath (path.js:39:11)
    at Object.join (path.js:1158:7)
    at Builder._unsymlinkModules (app/node_modules/ember-electron/lib/models/builder.js:55:26)
    at Builder.copyToOutputPath (app/node_modules/ember-electron/lib/models/builder.js:42:10)
    at Promise.resolve.then (app/node_modules/ember-cli/lib/models/builder.js:118:24)
...
```

It seems to be complaining about the `outputPath` being undefined, but specifying `--output-path` on the command line had no effect.

With some help on the Slack channel, I discovered that the problem is here: https://github.com/felixrieseberg/ember-electron/blob/159415a248014cd53fc3612da833939056252f9a/lib/commands/electron.js#L18-L47 .  If I change it to call `_super` at the end of the function, it works.  It appears that `availableOptions` needs to be specified prior to the base class constructor.

It appears that this commit is the culprit (and is therefore isolated to 2.9.0): https://github.com/felixrieseberg/ember-electron/commit/2b2047b468008dd27920b828bbc29bd197b79ce6

I'm assuming it must work in other versions of `ember-cli`, but I don't see anything in its changelog that might have impacted it.

jacobq:
I'm hoping this is all that is needed to fix #362 and that it doesn't cause other problems. Would someone else mind taking a look and commenting?

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -44,6 +44,7 @@ module.exports = {
     'comma-dangle': ['error', 'always-multiline'],
     'ember/no-const-outside-module-scope': 'off',
     'ember/no-direct-property-access': 'off',
+    'ember/avoid-leaking-state-in-ember-objects': 'warn', // TODO: Remove after upgrading eslint >= 4.2.0
     'ember/require-access-in-comments': 'off',
     'newline-before-return': 'error',
     'no-console': 'off',
@@ -77,6 +78,7 @@ module.exports = {
       plugins: ['node'],
       rules: Object.assign({}, require('eslint-plugin-node').configs.recommended.rules, {
         // add your custom rules and overrides for node files here
+        "ember/avoid-leaking-state-in-ember-objects": "off",
       })
     }
   ]
--- a/lib/commands/assemble.js
+++ b/lib/commands/assemble.js
@@ -10,15 +10,12 @@ module.exports = Command.extend({
   name: 'electron:assemble',
   description: `Assembles your app and places it into the output path (${defaultOut} by default).`,
 
-  init() {
-    this._super(...arguments);
-    this.availableOptions = [
-      { name: 'environment', type: String,  default: 'development', aliases: ['e', { 'dev': 'development' }, { 'prod': 'production' }], description: 'Possible values are "development", "production", and "test".' },
-      { name: 'platform',    type: String,  default: process.platform, aliases: ['p'] },
-      { name: 'build-path',  type: 'Path', aliases: ['i'], description: 'Reuse the build previously built via "ember electron:build" at this path rather than rebuilding.' },
-      { name: 'output-path', type: 'Path',  default: defaultOut, aliases: ['o'] },
-    ];
-  },
+  availableOptions: [
+    { name: 'environment', type: String,  default: 'development', aliases: ['e', { 'dev': 'development' }, { 'prod': 'production' }], description: 'Possible values are "development", "production", and "test".' },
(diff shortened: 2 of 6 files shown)


jacobq:
Ug, OK, so I'm already seeing that it's not this easy because one can't usually reference properties of `this` before calling `this._super`....

bendemboski:
@jacobq why don't we just go back to defining them as properties in the command class definition? That's how all the core `ember-cli` commands do it (e.g. [serve](https://github.com/ember-cli/ember-cli/blob/master/lib/commands/serve.js#L22-L42)), and also how `ember-exam`, a very widely used addon that supplies its own commands [does it](https://github.com/trentmwillis/ember-exam/blob/master/lib/commands/exam.js#L35-L41).

That's the idiomatic way to do it, so I don't think we should be making our commands look different from all the other ones out there because of a linting rule. I think we should either disable the rule using inline comments, or in the `overrides` section of `.eslintrc.js`.

jacobq:
@bendemboski I'm all for idiomatic but am confused at the apparent discrepancy. It seems like there are conflicting messages: i.e. "don't do X" & "doing X is the standard way". Is the reason it's OK in these cases because the values are immutable (at least in intention)?

If it were a general linter rule it would be easier for me to dismiss without further thought, but it's from the ember plugin, which is maintained by the `ember-cli` team:
https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/avoid-leaking-state-in-ember-objects.md#description

bendemboski:
Well, those rules are really meant for "Ember" code, i.e. code that runs in the browser on top of Ember, not code that runs in node on top of `ember-cli`. You'll see that neither `ember-cli` itself nor `ember-exam` use `eslint-plugin-ember`. It might make sense for us to remove that from `ember-electron` since we have very little code that runs in the browser, but regardless I don't think we should be writing our `ember-cli` code in non-idiomatic ways to accomodate eslint rules that weren't really meant to apply to it,

jacobq:
> Well, those rules are really meant for "Ember" code

This makes sense to me now. If I have time later today I'll update the commits in this PR unless someone beats me to it.

felixrieseberg:
> I think we should either disable the rule using inline comments, or in the overrides section of .eslintrc.js.

Strong agreement with that, let's cut down anything that makes it harder for you/us to keep this addon alive.

jacobq:
@felixrieseberg can you tell what is going wrong with the AppVeyor test [here](https://ci.appveyor.com/project/felixrieseberg/ember-electron/build/1.0.293#L790)?
I think it is a fluke so am planning to merge this anyway but wanted to get your opinion. I noticed in one of my private projects that I had to increase the browser connection timeout (to about 2 minutes) to ensure that it would run reliably in AppVeyor.

```
not ok 1 Electron - [undefined ms] - error
    ---
        message: >
            Error: Browser exited unexpectedly
            
        Log: |
            { type: 'error', text: 'Error: Browser exited unexpectedly' }
    ...
1..1
# tests 1
# pass  0
# skip  0
# fail  1
Testem finished with non-zero exit code. Tests failed.
```

PR #368, opened 2018-10-05
Fix for testing against electron 3 on windows

bendemboski:
https://github.com/electron/electron/pull/13039 caused our tests to fail on windows because we pass URL arguments. So, we need to add a '--' to the arguments to work around the issue.

### DIFF
--- a/blueprints/ember-electron/index.js
+++ b/blueprints/ember-electron/index.js
@@ -66,18 +66,10 @@ module.exports = class EmberElectronBlueprint extends Blueprint {
 
     logger.startProgress('Installing electron build tools');
 
-    return npmInstall.run({
-        // FIXME: Something seems to break in Windows with electron v3.0.0
-        // This is a hack to make electron-forge install electron 2.x instead of 'latest' (3.x) into the project
-        // See https://github.com/electron-userland/electron-forge/blob/5.x/src/api/import.js#L179
-        save: true,
-        verbose: false,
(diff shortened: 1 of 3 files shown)


bendemboski:
@jacobq fixed!

jacobq:
Care to also take out [the work-around](https://github.com/felixrieseberg/ember-electron/commit/5ad18658328462801b1db74b611462c5b178333f#diff-159eef82858485cb111c7e0a6db5d64eR70)?

jacobq:
BTW, does the `--` approach work with older versions and just not new ones?

bendemboski:
Yeah, good call on pulling the workaround -- just pushed a commit.

Adding the extra argument works fine with all versions. In the older versions it had a blacklist of command-line arguments that would cause it to fail to launch, but just a `--` wouldn't trigger anything, so would be left alone. The 3.0 change is to remove the explicit blacklist, and disallow arguments following a URL argument (which opens some security vulnerability related to protocol handling on windows) unless preceded by a `--` argument (which I guess blocks the vulnerability somehow). So either way, the arguments (including the `--`) will just get passed trough as-is, and `test-main.js` will parse them correctly, ignoring the `--`.

bendemboski:
@felixrieseberg this PR gets us electron 3.0 compatibility. Can you cut a release when you get the chance?

felixrieseberg:
Of course!

PR #381, opened 2018-12-28
fix: apply included function with arguments instead of app

makepanic:
This is the "recommended" way as seen in https://ember-cli.com/extending/#addon-entry-point .

I'm opening this PR as `this._super.included(app)` it causes some issues with ember-decorators and the next ember-cli-typescript version. 
This will prevent others from having to dive in on why this happens if you try to port your addon to use these addons.


jacobq:
@makepanic First of all, thanks for the PR!

>  ...it causes some issues with ember-decorators and the next ember-cli-typescript version.

Could you please elaborate? I am using the latest of both of those with `ember-electron` and have not seen any problems. Is it something very specific?

makepanic:
> @makepanic First of all, thanks for the PR!
> 
> > ...it causes some issues with ember-decorators and the next ember-cli-typescript version.
> 
> Could you please elaborate? I am using the latest of both of those with `ember-electron` and have not seen any problems. Is it something very specific?

Do you use the [ember-cli-typescript 2.0 rc](https://github.com/typed-ember/ember-cli-typescript/pull/333)? 
I've found that somehow the babel parser with ember-cli-typescript@2 isn't initialized with the decorators plugins if the addon calls the included "super" with `this._super.included(app);`:

```
[ 'classProperties',
  'classPrivateProperties',
  'typescript',
  'objectRestSpread',
  'asyncGenerators',
  'jsonStrings',
  'optionalCatchBinding' ]
```

This causes babel to throw as it sees decorators but can't parse them.
If one uses the arguments call (`this._super.included.apply(this, arguments);`) then the babel parser has all the expected plugins.

My guess is that some kind of scope mismatch happens. I could also open an issue with ember-cli-typescript as it's potentially also a bug upstream.

> So, @makepanic I think your code change here makes sense, and can you try moving loader.js from dependencies to devDependencies in package.json to see if that fixes the build?

I've moved it and rebased the PR again.

bendemboski:
Thanks for the contrib @makepanic!

PR #383, opened 2019-03-10
Bump electron-forge to 5.2.4

mnutt:
This upgrades `electron-forge` to the latest version, which I believe will fix this issue on MacOS:

https://github.com/electron-userland/electron-packager/issues/927

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -70,25 +70,38 @@
       "dev": true
     },
     "@octokit/rest": {
-      "version": "14.0.9",
-      "resolved": "https://registry.npmjs.org/@octokit/rest/-/rest-14.0.9.tgz",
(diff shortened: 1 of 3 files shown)


bendemboski:
Hey @jacobq I know #358 is something we'd still love to get back to, but since this is green right now, any objection to merging?

jacobq:
@bendemboski Sorry, not sure how I missed this. I have no objections. In fact, I have been using yarn selective resolutions to force this behavior in my projects for a while now so that I can use Electron versions newer than the latest published `electron-prebuilt-compile` package ([like this](https://github.com/electron-userland/electron-prebuilt-compile/issues/77#issuecomment-480881828)).

PR #389, opened 2019-04-23
Please publish

AlexandreTCF:
Please publish npm package:

`npm publish`

https://docs.npmjs.com/cli/publish

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ember-electron",
-  "version": "2.10.0",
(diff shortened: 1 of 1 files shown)


jacobq:
Hi @AlexandreTCF sorry for the delay. I think there's supposed to be an automatic release process, but perhaps there is something wrong with it or there needs to be some kind of trigger. I opened https://github.com/felixrieseberg/ember-electron/issues/391 to track this.

Issue #385, opened 2019-04-08
Incompatible with Electron 5.x


jacobq:
We currently rely on [`protocol.registerStandardSchemes`](https://electronjs.org/docs/api/protocol#protocolregisterstandardschemesschemes-options) here:

https://github.com/felixrieseberg/ember-electron/blob/dd9227aa5913e46658718b40143f6630edab996e/blueprints/ember-electron/files/ember-electron/main.js#L9

Starting in electron 5.x [this API is being removed](https://github.com/electron/electron/blob/8cf15cc9310ca5983f15e511c3ecdac8e75e225d/docs/api/breaking-changes.md#privileged-schemes-registration), and it looks like we'll need to replace the above call with something like:

```js
protocol.registerStandardSchemes({
  scheme: 'serve',
  privileges: {
    secure: true,
  }
});
```

pichfl:
We could also consider something like https://github.com/zeit/serve which might make serving the Ember app a bit more standardized. I know that the whole `serve` thing was my idea and it still works, but I still wonder if it was and is a good solution.

jacobq:
@pichfl I would be in favor of making things easier / less surprising for newcomers to learn. However, I'm not sure I understand what the pros and cons of each direction are. I would guess that most users don't even notice the custom protocol/scheme and that if they do they're able to see what's going on pretty quickly. (It's main goal is to facilitate hassle-free loading of assets without using `file://` while still passing other requests to the Ember app, right?)

Would other solutions (like the one you linked) handle the problem differently or just help reduce the size of the code base that needs to be maintained by the `ember-electron` team?

PR #390, opened 2019-04-24
fix: support Electron 5 (ref #385)

jacobq:
Use `protocol.registerSchemesAsPrivileged`, if available, since
`protocol.registerStandardSchemes` was removed in Electron 5.0.0
See https://github.com/electron/electron/blob/master/docs/api/breaking-changes.md#privileged-schemes-registration

Note: blueprint changed, so users will have to update their `ember-electron/main.js` file as well (or re-run `ember install ember-electron` to overwrite).

### DIFF
--- a/blueprints/ember-electron/files/ember-electron/main.js
+++ b/blueprints/ember-electron/files/ember-electron/main.js
@@ -6,7 +6,20 @@ const protocolServe = require('electron-protocol-serve');
 let mainWindow = null;
 
 // Registering a protocol & schema to serve our Ember application
-protocol.registerStandardSchemes(['serve'], { secure: true });
+if (typeof protocol.registerSchemesAsPrivileged === 'function') {
+  // Available in Electron >= 5
+  protocol.registerSchemesAsPrivileged([{
+    scheme: 'serve',
+    privileges: {
+      secure: true,
+      standard: true
+    }
+  }]);
+}
+else {
+  // For compatibility with Electron < 5
+  protocol.registerStandardSchemes(['serve'], { secure: true });
+}
 protocolServe({
   cwd: join(__dirname || resolve(dirname('')), '..', 'ember'),
   app,
--- a/lib/resources/shim-head.js
+++ b/lib/resources/shim-head.js
@@ -4,7 +4,18 @@
 
   // On linux the renderer process doesn't inherit the main process'
   // environment, so we need to fall back to using the remote module.
-  const serveIndex = process.env.ELECTRON_PROTOCOL_SERVE_INDEX || require('electron').remote.process.env.ELECTRON_PROTOCOL_SERVE_INDEX;
+  let serveIndex;
+  // If nodeIntegration is disabled, this will throw and serveIndex will remain
+  // undefined, which is fine because we only need it to fix module search paths
+  // that aren't relevant if node integration is disabled.
+  try {
+    serveIndex = process.env.ELECTRON_PROTOCOL_SERVE_INDEX || require('electron').remote.process.env.ELECTRON_PROTOCOL_SERVE_INDEX;
+  } catch (e) {
+    // When nodeIntegration is false we expect process to be undefined. Don't swallow the exception if something else is wrong
+    if (e.message !== "process is not defined") {
+      throw e;
+    }
+  }
 
   if (serveIndex && window.location.protocol !== 'file:') {
     // Using electron-protocol-serve to load index.html via a 'serve:' URL
--- a/lib/test-support/test-main.js
+++ b/lib/test-support/test-main.js
@@ -27,7 +27,18 @@ let emberAppDir = resolve(dirname(indexPath), '..');
 let relPath = relative(emberAppDir, indexPath);
 const emberAppLocation = `serve://dist/${relPath}${indexQuery}`;
 
-protocol.registerStandardSchemes(['serve'], { secure: true });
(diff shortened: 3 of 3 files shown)


jacobq:
Oops, just caught a mistake -- will fix momentarily:

```
A JavaScript error occurred in the main process
Uncaught Exception:
Error: Argument must be an array of custom schemes.
    at Object.<anonymous> (/path/to/app/electron-out/project/ember-electron/main.js:38:12)
    at Module._compile (internal/modules/cjs/loader.js:808:30)
    at Object.require.extensions.(anonymous function) [as .js] (/path/to/app/node_modules/electron-compile/lib/require-hook.js:77:14)
    at Module.load (internal/modules/cjs/loader.js:677:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:609:12)
    at Function.Module._load (internal/modules/cjs/loader.js:601:3)
    at Module.require (internal/modules/cjs/loader.js:715:19)
    at init (/path/to/app/node_modules/electron-compile/lib/config-parser.js:294:16)
    at main (/path/to/app/node_modules/electron-prebuilt-compile/lib/es6-init.js:58:29)
    at Object.<anonymous> (/path/to/app/node_modules/electron-prebuilt-compile/lib/es6-init.js:61:1)

```

jacobq:
Although my app now builds with this (and some other forks/hacks for `node-abi` and `electron-forge`) there's still a problem. Renders all white and shows error in console:

image

bendemboski:
Oh yeah, the process not defined issue has probably always been around with `nodeIntegration: false` I would think -- we should add some kind of check (although I'm not totally sure how to do that when checking for node built-ins...might have to be a try...catch block?)

bendemboski:
(but I agree with enabling nodeIntegration by default because I suspect the vast vast majority of ember-electron projects need node integration)

blueprints/ember-electron/files/ember-electron/main.js
```diff
@@ -32,6 +45,9 @@ app.on('ready', () => {
   mainWindow = new BrowserWindow({
     width: 800,
     height: 600,
+    webPreferences: {
+      nodeIntegration: true,
```
jacobq: It would be nice not to require `nodeIntegration` to be enabled in order to work. See https://github.com/felixrieseberg/ember-electron/pull/390#issuecomment-486440096

jacobq:
> (but I agree with enabling nodeIntegration by default because I suspect the vast vast majority of ember-electron projects need node integration)

I think it's a reasonable stop-gap to enable it by default for now because in the past we have assumed that it was enabled (otherwise :boom:). However, I don't know of a technical reason why we should require it, yet I can think of some classes of applications that would benefit from being able to disable it. So I think we should think about changing this in the future.

bendemboski:
I'm proposing we fix it right now. I think it's this easy:

```javascript
let serveIndex;
// If nodeIntegration is disabled, this will throw and serveIndex will remain
// undefined, which is fine because we only need it to fix module search paths
// that aren't relevant if node integration is disabled.
try {
  serveIndex = process.env.ELECTRON_PROTOCOL_SERVE_INDEX || require('electron').remote.process.env.ELECTRON_PROTOCOL_SERVE_INDEX;
} catch (e) {}
```

jacobq:
> I'm proposing we fix it right now. I think it's this easy:

OK. I hadn't poked around at this code yet so was a bit trepidatious. Is this the line that needs to be updated:

https://github.com/felixrieseberg/ember-electron/blob/master/lib/resources/shim-head.js#L7

bendemboski:
Yeah, exactly. All that code is just meant to set up node module search paths correctly, so it's totally safe to wrap in the equivalent of an `if (nodeIntegration)` block (and since there's already the truthy check for `serveIndex` below, just using a try...catch block to make sure that if node integration is disabled, `serveIndex` remains falsey should do the trick).

jacobq:
It's not really our fault, but I think it's worth noting that since I am using node integration in my app, tests failed to run with the default `lib/test-support/test-main.js` on Electron 5:

https://github.com/felixrieseberg/ember-electron/blob/069ca9df6f710006876cb4b2b270bc8bdbf213a9/lib/test-support/test-main.js#L64-L66

If others run into this it might be worth noting in the FAQ (if it's not already somewhere in the docs...I was too lazy to search very much).

devinus:
I'm not sure how this is supposed to work or even how you tested this while `ember-electron` still depends on `electron-prebuilt-compile` (which won't release a v5).

jacobq:
@devinus I am using [my own fork for electron-prebuilt-compile](https://www.npmjs.com/package/@jacobq/electron-prebuilt-compile) (see https://github.com/electron-userland/electron-prebuilt-compile/issues/77#issuecomment-480881828). It's a hack, admittedly, but it works well enough for my purposes. I I justify this hack to myself by thinking that once electron-forge v6.0.0 is released/stable we shouldn't need electron-prebuilt-compile any more and it can go away.

devinus:
@jacobq Gotcha. How do you get around electron-forge looking for the electron-prebuilt-compile package specifically?

jacobq:
> @jacobq Gotcha. How do you get around electron-forge looking for the electron-prebuilt-compile package specifically?

Same trick: used my own fork. e.g.

`"electron-forge": "https://registry.npmjs.org/@jacobq/electron-forge/-/electron-forge-5.2.4-jrq.6.tgz",`

PR #400, opened 2019-08-23
 chore(deps): run ember-cli-update to 3.12.x

bendemboski:
This appears to fix the weird acceptance test failures that were happening. Since the acceptance tests create new Ember projects, they necessarily have floating dependencies, so I'm guessing some dependency update messed something up. But moving to the latest ember-cli seems to work. So, yay?


jacobq:
Sorry I missed this (have been in embedded-development land for the past couple months). Thanks for giving `ember-electron` some ❤️

PR #401, opened 2019-08-24
fix: enable node integration in tests

bendemboski:
So Electron >=5 behaves the same as Electron <5.


bendemboski:
The background here is that `nodeIntegration` was `true` by default before Electron 5, so this smooths the upgrade path. It makes sense to require users to opt in to node integration when creating browser windows in their applications, but the likelihood that none of their tests rely on node integration is quite low (most Electron apps use node in the renderer processes after all), so turning it on for tests seem like the right thing to do regardless of compatibility stories.

PR #404, opened 2019-11-02
docs: Use ember-cli-addon-docs for docs

rwwagner90:
We might want to also move the current homepage info to the index of the docs page, but this is a start.


rwwagner90:
@bendemboski here is a start on addon docs

rwwagner90:
@bendemboski do you know how the current site is deployed and hosted? I think we should make the current site be the index page of addon docs, then we should get addon docs deployed.

jacobq:
(:wave:, Jacob the lurker here)
Thanks for working on this! 

> do you know how the current site is deployed and hosted?

I wasn't the one who set it up, but I believe it's hosted with a typical GitHub Pages setup (you can access it via http://adopted-ember-addons.github.io/ember-electron and that appears to match the [`gh-pages` branch](https://github.com/adopted-ember-addons/ember-electron/tree/gh-pages)). It just happens to use a custom DNS name through [js.org](https://js.org/) (which shouldn't need to be changed). I believe you can deploy just by pushing the file to the gh-pages branch.

PR #405, opened 2019-11-07
docs: Change links to docs, port index page

rwwagner90:
@bendemboski @jacobq I need to figure out how to port the styles for the index page over correctly. They were reliant on the font being 62.5%, so everything is thrown off a lot. This gets all the pages in the right spot though, and updates links, so we can deploy the docs to gh-pages.


pichfl:
@rwwagner90thanks for taking this up.

I made that index page on my flight to EmberConf a few years ago. If you need any resources, I can provide the design files if necessary.

rwwagner90:
@pichfl yeah replacing with px was my next thing to try. I think after that, the remaining issues are the content is not staying in the div for some reason, and the buttons over the curved part are being hidden under it. That might just be z-index. Have not looked yet.

rwwagner90:
@pichfl got the styles ported over.

@bendemboski @jacobq once we get this in, we should be able to deploy addon docs to the gh-pages branch. Would it be possible for me to get permissions to this repo, so I could follow up with that step?

PR #410, opened 2019-11-16
chore: Make ESLint config more standard

rwwagner90:


.eslintrc.js
```diff
@@ -26,14 +26,9 @@ module.exports = {
...
   rules: {
-    'ember/no-const-outside-module-scope': 'off',
-    'ember/no-direct-property-access': 'off',
-    'ember/avoid-leaking-state-in-ember-objects': 'warn', // TODO: Remove after upgrading eslint >= 4.2.0
     'ember/no-jquery': 'error',
-    'ember/require-access-in-comments': 'off',
-    'newline-before-return': 'error',
     'no-console': 'off',
-    'one-var': 'off',
+    'no-prototype-builtins': 'off'
```
bendemboski: Why are we disabling this rule?

.eslintrc.js
```diff
@@ -65,6 +61,7 @@ module.exports = {
       rules: Object.assign({}, require('eslint-plugin-node').configs.recommended.rules, {
         // add your custom rules and overrides for node files here
         'ember/avoid-leaking-state-in-ember-objects': 'off',
+        'node/no-unpublished-require': 'off'
```
bendemboski: Is this just for `electron`? Could we configure it explicitly with `"allowModules": [],` instead of disabling the rule entirely?

package.json
```diff
@@ -115,10 +115,8 @@
     "ember-source": "~3.14.1",
     "ember-source-channel-url": "^2.0.1",
     "ember-try": "^1.2.1",
-    "eslint": "^4.18.2",
-    "eslint-config-ember": "^0.3.0",
-    "eslint-plugin-ember": "^7.1.0",
-    "eslint-plugin-ember-suave": "^1.0.0",
+    "eslint": "^6.6.0",
```
bendemboski: Oh, I missed before that we have `eslint` as a direct dependency. How about we get rid of `ember-cli-eslint` then so we don't have two versions kicking around?

.eslintrc.js
rwwagner90: There was something violating it, so I disabled it.

bendemboski: Just one? Can we just do a line comment to disable it then? There are security implications to this rule, and while I'm not certain they apply to Electron or not, with all the various security risks that have to be carefully monitored/managed in Electron, I think it would be better to keep the rule on.

package.json
rwwagner90: I think the opposite it what we want, but I went ahead and remove eslint, so we can closer align to Ember projects.

bendemboski: Oh, okay. Everything I've heard indicates that `ember-cli-eslint` is going to be removed from the default blueprint now that we have `lint:js` in the travis config. But it hasn't yet, so I'm fine doing it this way.

.eslintrc.js
rwwagner90: Seems to somehow no longer be an issue

PR #412, opened 2019-11-16
chore: Fix vendor weirdness

bendemboski:
For some reason this script was in app/, but index.js was broccoli'ing app/ into vendor/. So move the script to vendor to make things less confusing.


package.json
```diff
@@ -41,12 +41,11 @@
   "scripts": {
     "lint:hbs": "ember-template-lint .",
     "lint:js": "eslint .",
-    "test": "yarn-or-npm run test-node-all && yarn-or-npm run lint",
```
rwwagner90: Rather than totally removing `yarn-or-npm run lint` shouldn't we change it to `yarn-or-npm run lint:js`?

bendemboski: I don't think so. We're kinda trying to be Ember-ish here, and in the Ember blueprint, `yarn test` runs `ember test` and `yarn lint:js` lints (and CI runs both) -- linting is separate from testing. That's what I was going for here -- make sense?

PR #418, opened 2019-11-21
Update to electron-forge 6 & simplify build pipeline/project structure

bendemboski:
This is a very significant, and breaking, change, but should yield very big improvements in developer ergonomics. The documentation explains in detail, but the two major changes are updating to `electron-forge` 6, and significantly simplifying the build pipeline:

* `electron-forge` 6 allows us to use much more modern Electron tooling, including updating Electron directly rather than relying on libraries like `electron-prebuilt-compile` to release
* the revamped build pipeline allows a much simpler mental model of Ember app vs. Electron app
* the revamped build pipeline is much faster at building/packaging/making as it removes a whole npm/yarn install step
* the revamped build pipeline exposes the Electron project as a standard `electron-forge` project, optionally allowing the direct use of all of its tooling (including `electron-forge publish`).
* `ember electron` uses the same live reload mechanism as `ember serve`, including hot reloading CSS


blueprints/ember-electron/files/testem-electron.js
```diff
@@ -1,6 +1,6 @@
 /* eslint-env node */
 module.exports = {
-  test_page: 'tests/index.html?hidepassed',
+  test_page: 'tests/index.html?hidepassed=1',
```
rwwagner90: Forgive my ignorance, but what does the `=1` do?

blueprints/ember-electron/index.js
```diff
@@ -27,122 +29,127 @@ module.exports = class EmberElectronBlueprint extends Blueprint {
...
 
-        logger.message(message);
-      });
+    this.ui.writeLine(chalk.green('Updating .travis.yml'));
+
+    try {
+      let contents = await readFile('.travis.yml');
+      let yawn = new YAWN(contents.toString());
+
+      let doc = yawn.json;
```
rwwagner90: We should definitely explore using the built in ember-cli methods for modifying files https://ember-cli.com/api/classes/Blueprint.html#method_insertIntoFile

blueprints/ember-electron/files/testem-electron.js
bendemboski: Oh, I should revisit that or add a comment. It doesn't actually make a difference in-browser, but some of the URL munging that we do in `test-runner.js` was resulting the the query string looking like `?hidepassed=&testemid=whatever` and apparently that was preventing QUnit from recognizing the `hidepassed` query param. Giving at value fixed it.

blueprints/ember-electron/index.js
bendemboski: Yeah, I switched to that on your suggestion for `.eslintignore`, but I'm more nervous about `.travis.yml`. Since the document has a structure that's so much more complex than an ignore file, I was thinking it would be better to use an AST parser rather than trying to handle all the various ways users might have modified their configs using text-only parsing.

package.json
```diff
@@ -120,6 +120,9 @@
     "tmp": "^0.1.0",
     "walk-sync": "^2.0.2"
   },
+  "peerDependencies": {
+    "ember-cli": "^3.0.0"
+  },
```
jacobq: I like this so much better than white-listing the linting rule :heart:

lib/tasks/assemble.js
```diff
@@ -65,9 +65,9 @@ class AssembleTask extends Task {
 
   pruneCommand() {
     if (shouldUseYarn(this.project.root)) {
-      return 'yarn install --production --no-bin-links';
```
jacobq: Do you remember why this was necessary before? I am guessing that it had to do with rebuilding native addons for electron.

bendemboski: I don't recall, but this whole task is deleted in a later commit, so probably not worth worrying about.

package.json
bendemboski: Yeah, totally. I used to be terrified of peerDependencies, but I've started to find more uses for them -- mostly around linting and keeping `ember-auto-import` happy 😄

blueprints/ember-electron/index.js
```diff
@@ -1,20 +1,22 @@
...
+const chalk = require('chalk');
+const {
+  electronProjectPath,
+  emberBuildDir,
+  emberTestBuildDir
+} = require('../../lib/utils/build-paths');
+const path = require('path');
+const denodeify = require('denodeify');
+const fs = require('fs');
+const readFile = denodeify(fs.readFile);
```
jacobq: Since this is running on node >= 10, what do you think about using the [native "fs Promises API"](https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_fs_promises_api) instead of using `denodeify(fs...)`? Also, for cases not covered by fsPromises what do you think about using the [native `util.promisify`](https://nodejs.org/dist/latest-v10.x/docs/api/util.html#util_util_promisify_original)? Seems like a good time to drop another dependency.

node-tests/integration/commands/electron-test.js
```diff
@@ -0,0 +1,138 @@
...
+    // throws an error
+    willInterruptProcess.capture(mockProcess);
+  });
+
+  afterEach(function() {
+    process.argv = processArgv;
+    willInterruptProcess.release();
+  });
+ 
+  it('works', async function() {
```
jacobq: Not to be hypocritically nit-picky (as I've contributed exactly zero lines of code to this PR), but the other test names are quite descriptive, so seeing a few "it works" tests surprised me. Is it standard/conventional to use a generic name like this for "smoke tests" or were these just overlooked?

lib/commands/test.js
```diff
@@ -1,24 +1,41 @@
...
+      name: 'reporter',
+      type: String,
+      aliases: ['r'],
+      description: 'Test reporter to use [tap|dot|xunit] (default: tap)',
+    },
+    { name: 'silent', type: Boolean, default: false, description: 'Suppress any output except for the test report' },
+    { name: 'testem-debug', type: String, description: 'File to write a debug log from testem' },
+    { name: 'test-page', type: String, description: 'Test page to invoke' },
+    { name: 'query', type: String, description: 'A query string to append to the test page URL.' },
+  ],
```
jacobq: This appears to be mostly (all?) duplicated from [`ember-cli/lib/commands/test.js`](https://github.com/ember-cli/ember-cli/blob/b0e01db47ae7e369d4acb937d0d580dae80e9d84/lib/commands/test.js#L22-L79). Is there an easy way we could import it from there instead?

lib/resources/shim-test-head.js
```diff
@@ -1,12 +1,11 @@
...
+// server -- see https://github.com/testem/testem/commit/4a51acc2fc0c3a23273fea838fd166b4691c2300.
+//
+// The testemId query param is added to the test URL by test-runner.js.
 //
 window.getTestemId = function() {
-  // FIXME: It should be safe to replace "\?" with "?" below due to context -- need to check though
-  // eslint-disable-next-line no-useless-escape
-  let match = window.location.search.match(/[\?&]testemId=([^\?&]+)/);
-
+  let match = window.location.search.match(/[?&]testemId=([^?&]+)/);
```
jacobq: Thank you for fixing my FIXME :D

lib/test-support/test-index.js
```diff
@@ -25,7 +22,7 @@ if (process.platform === 'win32') {
 
 let emberAppDir = resolve(dirname(indexPath), '..');
 let relPath = relative(emberAppDir, indexPath);
-const emberAppLocation = `serve://dist/${relPath}${indexQuery}`;
+let emberAppLocation = `serve://dist/${relPath}${indexQuery}`;
```
jacobq: Was this `const` changed to `let` just to match its neighbors? (Doesn't look like it get changed)

lib/test-support/test-runner.js
```diff
@@ -25,47 +25,54 @@ module.exports = {
...
     query: testPageQuery,
   } = url.parse(testPageUrl, true);
   let indexPath = path.resolve(emberAppDir, path.join.apply(null, testPagePath.split('/')));
   let indexObj = url.parse(fileUrl(indexPath));
   indexObj.query = testPageQuery;
+
+  // https://github.com/nodejs/node/issues/9500
+  for (let [ key, value ] of Object.entries(indexObj.query)) {
+    if ([ null, undefined, '' ].includes(value)) {
+      indexObj.query[key] = 'true';
```
jacobq: Just to make sure I understand correctly, we want to work around the behavior described in the URL mentioned, so we interpret any query string parameter that doesn't have an associated value (`=...` / right-hand side) as meaning "this parameter is present" and thus assign it a value of "true" (sort of like [boolean HTML attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#Boolean_Attributes)).

package.json
```diff
@@ -42,51 +42,33 @@
...
     "ember-inspector": "^3.12.5",
-    "execa": "^3.3.0",
     "file-url": "^3.0.0",
-    "fs-extra": "^8.1.0",
-    "globby": "^10.0.1",
-    "lodash": "^4.15.0",
-    "quick-temp": "^0.1.5",
+    "ncp": "^2.0.0",
+    "portfinder": "^1.0.25",
     "rimraf": "^3.0.0",
```
jacobq: I think `rimraf` can get moved to `devDependencies` as well, right?

blueprints/ember-electron/files/testem-electron.js
bendemboski: I fixed it in `test-runner.js` and removed the `=1`

blueprints/ember-electron/index.js
bendemboski: If you look at `engines` in `package.json`, we still support Node 8, so can't use promises directly. But I didn't realize promisify was rolled into Node in v8, so I'll def switch to that so we can drop the extra dependency!

node-tests/integration/commands/electron-test.js
bendemboski: I've seen this pattern pretty commonly -- when a test module is testing a very specific subject (in this case the `ember electron` command), an `it works` test that tests the basic functionality of the subject, followed by a bunch of more specific tests. We didn't have one of these before because we didn't have a basic functionality tests, only tests of specific conditions, which IMO is a bit of a testing faux pas.

bendemboski: (also, not hypocritical -- coding is only part of the story -- review and feedback is vital to making changes, especially big ones like this, so thank you for taking the time!)

tests/dummy/app/templates/docs/guides/upgrading.md
```diff
@@ -0,0 +1,126 @@
...
+
+If yours still looks like that, then you don't need to do anything. If you've customized it, then you'll need to migrate the content to `electron-app/tests/index.js`
+
+### resources-* folders
+
+If you don't have any files in any of the `ember-electron/resources` or `ember-electron/resources-*` folders, you can skip this section.
+
+`ember-electron` 2.x managed the `resources-*` folders to allow you to specify platform-specific resources. `electron-forge` does not support this functionality, and it was one of the main factors contributing to the complexity and slowness of `ember-electron` 2.x's build pipeline, so it's been removed in 3.x.
+
+So if you only have content in `ember-electron/resources/`, you can just copy the folder into `electron-app`. Note that `src/index.js` is now in a subfolder, unlike `main.js` was in 2.x, so if you are accessing resources from your main process using, e.g., `path.join(__dirname, 'resources')`, you'll have to update it to `path.join(__dirname, '..', 'resources')` (or you could put the `resources` folder in the `src/` directory if you're some kind of monster).
```
rwwagner90:
```suggestion
So if you only have content in `ember-electron/resources/`, you can copy the folder into `electron-app`. Note that `src/index.js` is now in a subfolder, unlike `main.js` was in 2.x, so if you are accessing resources from your main process using, e.g., `path.join(__dirname, 'resources')`, you'll have to update it to `path.join(__dirname, '..', 'resources')`. Alternatively, you could put the `resources` folder in the `src/` directory, but it is not recommended.
```

tests/dummy/app/templates/docs/guides/development-and-debugging.md
```diff
@@ -0,0 +1,68 @@
...
+$ ember electron --environment=production --- --some-flags --that-my-app --handles-as="process.argv"
+```
+
+## Ember Inspector and Devtron
+
+`ember electron` installs the Electron Inspector and Devtron directly into Electron, so they should be available if you open the dev tools. Note that if you're in a browser window without `nodeIntegration` enabled, they will not be installed.
+
+## Developing main process code
+
+`ember-electron` does not restart your Electron app when main process code changes, so after making a code change you need to quit and re-launch your application. If you just run `ember electron` each time and you aren't changing any Ember code, you'll have to unnecessarily wait through the Ember build process that `ember electron` automatically runs. To avoid this, you can
```
rwwagner90:
```suggestion
`ember-electron` does not restart your Electron app when main process code changes, so after making a code change you need to quit and re-launch your application. If you only run `ember electron` each time and you aren't changing any Ember code, you'll have to unnecessarily wait through the Ember build process that `ember electron` automatically runs. To avoid this, you can
```

tests/dummy/app/templates/docs/guides/upgrading.md
```diff
@@ -0,0 +1,126 @@
...
+ember install ember-electron
+```
+
+When prompted to overwrite `testem-electron.js`, choose `yes`.
+
+## Fix electron version
+
+Your `electron-app` project will initially have the most recent release of Electron installed. You were probably running an older version since it was very difficult to run newer than Electron 4.x with `ember-electron` 2.x. So if you don't want to deal with upgrading Electron in the middle of upgrading `ember-electron`, you should change your `electron-app/package.json`'s `electron` version to match the version of `electron-prebuilt-compile` (which would determine the Electron version you were running) that is in your root `package.json`. Don't forget to run `yarn`/`npm install` from `electron-app` afterwards!
+
+The good news is that now upgrading Electron just involves updating the version installed in the `electron-app` package -- no more waiting for `electron-prebuilt-compile` to release an update matching the latest Electron version!
```
rwwagner90:
```suggestion
The good news is that now upgrading Electron only involves updating the version installed in the `electron-app` package -- no more waiting for `electron-prebuilt-compile` to release an update matching the latest Electron version!
```

lib/commands/test.js
bendemboski: Good idea, I just pushed a commit to do that for this as well as the build command

lib/test-support/test-index.js
bendemboski: Heh, don't really want to get into the `const` vs. `let` debate, although I think this case is particularly ambiguous. I'd prefer to leave it as-is, if you'd like to disagree and discuss this particular case, I'm happy to, and if you'd like to have the stylistic discussion of `const` vs. `let` in general, hit me up on Discord 😄

jacobq:
> ...suggest we remove the bit about being a monster if you use an alternate directory placement.

After hearing so much about how it's hard for outsiders to get into the tech industry (e.g. from [Ellen Ullman](https://en.wikipedia.org/wiki/Ellen_Ullman)'s [story](https://www.amazon.com/Life-Code-Personal-History-Technology/dp/0374534519)) and considering that those whose native language is not English may not be familiar with the expression, I think you're probably right. Nevertheless, the disappearance of this kind of innocent lightheartedness still strikes me as unfortunate.

blueprints/ember-electron/index.js
jacobq: Do we need to keep support for node 8? It will reach end-of-life status in one month, and this is a semver major release.

lib/test-support/test-index.js
jacobq: No, no, it's fine. I wasn't trying to get into a debate. I was just a bit confused as to why it was changed. The commit said "get tests working" but this didn't appear to be needed to get tests working.

blueprints/ember-electron/index.js
jacobq: (Just saw that you updated CI to run on 8 so am marking conversation as resolved)

lib/test-support/test-runner.js
bendemboski: Yeah, it's a kinda annoying bug in the URL library. When you assign a search object to a parsed URL, any keys whose values are the empty string (or undefined/null I think) end up doing this:

```javascript
const URL = require('url');

let obj = URL.parse('http://localhost/index.html', true);
obj.query = { hidepassed: '' };
console.log(URL.format(obj)); // http://localhost/index.html?hidepassed=
```

there's probably another workaround where we assign the query string as an un-parsed string, then round-trip format/parse it and then assign the testem id, but seems like more code/possibly less clear than this.

lib/test-support/test-index.js
bendemboski: Yeah, fair, there was so much code change here that I tried to rebase things to make the commits sensible, but didn't go so far as to do the pain of resolving conflicts and splitting out changes I accidentally combined.

package.json
bendemboski: No, we use it in the electron-forge template which runs from the blueprint.

jacobq:
> Well, I'm gonna call "not my fault" on this one -- pre-existing error

Guilt by association: you edited something nearby so now you have to fix it :rofl:

tests/dummy/app/templates/docs/guides/upgrading.md
bendemboski: The language here is trying to emphasize that this is all you need to do as far as copying files if you don't have any platform-specific resources, but if you do have platform-specific resources, you need to do some other stuff.

Maybe

> So if you only have content in `ember-electron/resources/`, you can copy the folder into `electron-app` and ignore the other `ember-electron/resources-*` folders.

?

tests/dummy/app/templates/docs/guides/development-and-debugging.md
bendemboski: Maybe it's stylistic, but this actually reads less clear to me. What's your argument against `just` in this context?

tests/dummy/app/templates/docs/guides/upgrading.md
bendemboski: (same comment here as above)

tests/dummy/app/templates/docs/guides/upgrading.md
rwwagner90: Seems fine to me. 👍

tests/dummy/app/templates/docs/guides/development-and-debugging.md
rwwagner90: I feel like `only === just` in meaning here, as the thing you are trying to convey is you only run `ember electron`.

bendemboski:
> ...suggest we remove the bit about being a monster if you use an alternate directory placement.

🤷‍♂ okay. This doesn't strike me as a case that's going to really cause harm, and I believe there are soft benefits in keeping some levity and humor in our community environments, but don't feel strongly, so I'll acquiesce to the 2 against 1 vote! Unless y'all would be happy adding an emoji or something to clarify that it's a joke 😝?

Otherwise, how about:

> or you could even put the `resources` folder in the `src/` directory if you like, although many folks prefer to not mix their code with non-code resources

tests/dummy/app/templates/docs/guides/upgrading.md
rwwagner90: I'm not the one who started the war on "just", I know it gets flagged by a lot of tools as being discriminatory or something though.

bendemboski: Like is it an ESL thing? Can you link to any articles or anything?

To be blunt, I'm not inclined to follow advice that seems wrong to me if I don't have any understanding of the reasoning behind it. I'm certainly not going to die on this hill if you feel really strongly about it, but even after a little googling, I'm unable to get any sense of why `only` would be preferable to `just` here.

bendemboski:
Okay. Maybe I have more faith in "the community" and maybe that's super naive, but also I don't care that much, so I'll change it.

tests/dummy/app/templates/docs/guides/upgrading.md
rwwagner90: I think the argument is it has a negative connotation like "just do it, it's easy", "just make it work" etc. This case doesn't necessarily apply. I don't particularly care that much, only following what I have heard. There was a talk at EmberConf about it, and it sounds like whatever tool they use only looks for how many times you said "just", "guys", etc anything that could be referring to gender or implying that something is easy, when it is not easy for everyone.

bendemboski:
I think I've addressed all of the comments on this PR -- let me know if I've missed any. Before merging I want to port my big complex `ember-electron` app to this new version to see if it turns up anything we've missed. I think I can do that today, but certainly by Monday.

@jacobq @rwwagner90 anything you can think of that we should wait on before merging?

rwwagner90:
@bendemboski I think this looks good to me now. I'm running it in my app, and it all works!

jacobq:
> anything you can think of that we should wait on before merging?

I won't have a chance to migrate my main app to it until late next week, but I don't think we should hold up merging just for that. Perhaps it could merge and be a beta (3.0.0-beta.1?) release so there's a chance to make more breaking changes if major problems come to light afterward?

bendemboski:
Oh yeah, we're going to beta it for sure. I just want to do my migration before even doing that unless I get hung up and can't have it done by Monday. But yeah, definitely beta!

jacobq:
In case you rebase again, you may want to fix the typo in the commit message:
defualt -> default

jacobq:
Looks like the [Windows build failures](https://ci.appveyor.com/project/rwwagner90/ember-electron/builds/29078797#L146) are mostly/all due to hard-coded POSIX-style path separator. These should be easily fixable by replacing things like [`'electron-app/out'`](https://github.com/adopted-ember-addons/ember-electron/pull/418/files#diff-78dbf845b16f07f0a3a24bac84a76305R123) with `path.join('electron-app', 'out')`

jacobq:
> Maybe I have more faith in "the community" and maybe that's super naive...

The way to think about it that's been helpful to me is to realize that context is not static. What may be totally appropriate, clear, acceptable, etc. in the environment where it originated may later be terribly misconstrued. When I worked at 3M we were often encouraged to mentally run every electronic message we authored through the filter of "If you wouldn't want to see this in the headlines of a newspaper [or defend it in court] then don't write it." and there's definitely some wisdom in that. The emberjs developer community of today may later expand/change, etc. meanwhile history as recorded by version control systems doesn't have the flexibility to clarify or explain itself.

rwwagner90:
I think we're ready to merge this and get a beta out. Any objections? Once this is merged, I think we should try to add https://github.com/rwjblue/create-rwjblue-release-it-setup to setup releases and changelogs for us.

bendemboski:
Sure, go for it!

jacobq:
> Once this is merged, I think we should try to add https://github.com/rwjblue/create-rwjblue-release-it-setup to setup releases and changelogs for us.

Fine by me. I thought Felix had set up something like that already though (semantic-release bot).... is that no longer working  / now obsolete?

rwwagner90:
@jacobq we removed it because we didn't like it releasing for us. We wanted to be able to decide when to bump versions and release, especially with releasing betas like this.

PR #421, opened 2019-11-25
Docs after release

rwwagner90:
I think this should work, and ensures we don't forget to update docs.

### DIFF
--- a/.release-it.json
+++ b/.release-it.json
@@ -1,0 +1,5 @@
+{
+  "hooks": {
+    "after:release": "ember deploy production"
+  }
+}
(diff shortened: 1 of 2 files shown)


.release-it.json
```diff
@@ -0,0 +1,5 @@
+{
+  "hooks": {
+    "after:release": "ember deploy production"
```
jacobq: I am confused by [the docs](https://github.com/release-it/release-it#hooks). What is the difference between `afterRelease` and `after:release`?

rwwagner90: I honestly have no idea. Was just going to try this out, and if it didn't work we can tweak 😛

PR #422, opened 2019-11-26
doc: remove duplicate `docs/` path segment

jacobq:
I was browsing https://adopted-ember-addons.github.io/ember-electron/versions/v3.0.0-beta.0/docs/#documentation and noticed that most of the links go to a blank/bogus page. Looks like this is because the path is relative and this file is in `docs` already, so resultant URLs contain `docs/docs`.

### DIFF
--- a/tests/dummy/app/templates/docs/index.md
+++ b/tests/dummy/app/templates/docs/index.md
@@ -20,18 +20,18 @@ To see a real world example, check out [Ghost Desktop](https://github.com/trygho
 ## Documentation
 
 ### Basics
-- [Installation](docs/guides/installation)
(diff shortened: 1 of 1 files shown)


jacobq:
> @jacobq can we remove all the documentation links instead? The links are all in the side nav now.

That's fine with me. I was just going for the quickest fix possible. (Noticed this last night while I was migrating an app to try the beta. Unfortunately I'm at work now and can't contribute a PR at the moment. Feel free to close this and fix it however you like.)

rwwagner90:
Went ahead and removed them. Thanks for catching this!

PR #426, opened 2019-11-26
doc: suggest including @3 suffix

jacobq:
Since `ember install ember-electron` will use the `latest` npm tag by default, these instructions won't work until that tag points to 3.x. Since these instructions are specifically for upgrading from 2.x to 3.x specifying `ember install ember-electron@3` seems like a reasonable solution.

### DIFF
--- a/tests/dummy/app/templates/docs/guides/upgrading.md
+++ b/tests/dummy/app/templates/docs/guides/upgrading.md
@@ -123,4 +123,4 @@ If you need to exclude files for other platforms from your packaged build, you c
 
 ### code/files/etc
 
-Anything else that was in the `ember-electron` in 2.x should be "just files" -- `.js` files `require`d from the main process or `requireNode`d from the Ember app, or other files accessed via the filesystem APIs. So these can be migrated into `electron-app` however seems best, updating the references to their paths in other source files as needed.
+Anything else that was in the `ember-electron` folder in 2.x should be "just files" -- `.js` files `require`d from the main process or `requireNode`d from the Ember app, or other files accessed via the filesystem APIs. So these can be migrated into `electron-app` however seems best, updating the references to their paths in other source files as needed.
--- a/tests/dummy/app/templates/docs/guides/upgrading.md
+++ b/tests/dummy/app/templates/docs/guides/upgrading.md
@@ -6,7 +6,7 @@ While upgrading from 2.x to 3.x should not be very time-consuming in most cases,
 
 ```sh
 yarn-or-npm remove ember-electron
-ember install ember-electron
+ember install ember-electron@3
 ```
 
 When prompted to overwrite `testem-electron.js`, choose `yes`.


jacobq:
Oh dear, I didn't realize 3 was someone's user name. Sorry for the accidental shout-out ;) Have a good day

tests/dummy/app/templates/docs/guides/upgrading.md
```diff
@@ -6,7 +6,7 @@ While upgrading from 2.x to 3.x should not be very time-consuming in most cases,
 
 ```sh
 yarn-or-npm remove ember-electron
-ember install ember-electron
+ember install ember-electron@3
```
rwwagner90: Should we maybe do `@beta` instead?

tests/dummy/app/templates/docs/guides/upgrading.md
```diff
@@ -123,4 +123,4 @@ If you need to exclude files for other platforms from your packaged build, you c
 
 ### code/files/etc
 
-Anything else that was in the `ember-electron` in 2.x should be "just files" -- `.js` files `require`d from the main process or `requireNode`d from the Ember app, or other files accessed via the filesystem APIs. So these can be migrated into `electron-app` however seems best, updating the references to their paths in other source files as needed.
\ No newline at end of file
+Anything else that was in the `ember-electron` in 2.x should be "just files" -- `.js` files `require`d from the main process or `requireNode`d from the Ember app, or other files accessed via the filesystem APIs. So these can be migrated into `electron-app` however seems best, updating the references to their paths in other source files as needed.
```
rwwagner90: Not sure what actually changed in this part, but can we fix this while we are here?

```suggestion
Anything else that was in the `ember-electron` folder in 2.x should be "just files" -- `.js` files `require`d from the main process or `requireNode`d from the Ember app, or other files accessed via the filesystem APIs. So these can be migrated into `electron-app` however seems best, updating the references to their paths in other source files as needed.
```

tests/dummy/app/templates/docs/guides/upgrading.md
jacobq: Well, then if we come out with `4.0.0-beta.1` later it will be wrong. Doesn't `@3` work? (I didn't actually test)

rwwagner90: I don't think we would keep the instructions once a new beta was out. They replaced the old 1.x to 2.x upgrade instructions. We could split them out and have separate upgrading pages for 1 -> 2, and 2-> 3.

jacobq:
Closing since I just realized that using this suffix doesn't work right now (since 3.x is still in pre-release).

rwwagner90:
@jacobq you can use `@beta` to install the beta

jacobq:
> @jacobq you can use `@beta` to install the beta

Yes, that's what I did. The issue/PR is about what the docs say/assume. They say to implicitly install `@latest` / assume there is a 3.x release, and that is OK as long as beta users know to append `@beta` until 3.0.0 is released.

PR #432, opened 2019-12-07
fix: use a single version of electron-forge

### LINKED ISSUE #431: Issues with electron-forge versions
I ran a `yarn upgrade` in my ember project, and unknowingly broke my build, since electron-forge is bundled with ember-electron, and had just released a new version. I spoke to @bendemboski about it a bit, and he has some ideas to try to consolidate versions of electron-forge. We'll need to figure out a solution, so this does not happen in the future.

bendemboski:
We have two versions of electron-forge kicking around. ember-electron has it as a `dependency` so it can use it from the blueprint to create the Electron project using electron-forge's `init` command. When electron-forge creates a project, it installs itself in that project, so we end up with another version in `electron-app/package.json`.

When `ember-electron` was calling electron forge commands from `ember electron:*` commands, it was just `require()`ing electron-forge, which would get it the version in `/node_modules`. This could be a different version from the one in `electron-app/node_modules`, so the user could get different behavior running, e.g. `ember electron:make` vs. `cd electron-app && yarn electron-forge make`, breaking the simple mental model.

Also, recently electron-forge released a new beta version that changed the maker interface slighty, so the makers installed in `electron-app/` could be out of sync with the `electron-forge` version that the `ember electron` commands use. This is probably only a problem while electron-forge is in beta, but is also addressed by this change.

Now, other than the initial Electron project creation that happens from the blueprint, we only use the electron-forge in `electron-app`, which puts it right next to the makes in `package.json` allowing the user to easily see and manage the version without any hidden complexity.

Fixes #431


lib/utils/forge-core.js
```diff
@@ -0,0 +1,14 @@
+const { electronProjectPath } = require('./build-paths');
+
+try {
+  module.exports = require(require.resolve('@electron-forge/core', {
+    paths: [ electronProjectPath ]
+  }));
+} catch (e) {
+  throw new Error([
```
jacobq: Just curious, why `new Error(...)` instead of [shorthand `Error(...)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error)?

bendemboski: I'm not sure what you're talking about, I don't see any description of a shorthand form in that link. Error is a class and I instantiate classes with new 🤷

jacobq:
Summarizing to check my understanding:

1. Motivation: `ember-electron` should avoid using its own copy (local `node_modules`) of any `electron-forge` packages (`@electron-forge/*`) to avoid discrepancies (e.g. version mismatch) / improve consistency (e.g. effectively the same behavior with `ember electron:...` as with equivalent forge commands in `electron-app`). 
2. New file `lib/utils/forge-core.js` wraps calls to `require('@electron-forge/core')` and [forces](https://nodejs.org/api/modules.html#modules_require_resolve_request_options) use of version in `electron-app/node_modules/` (or [`GLOBAL_FOLDERS`](https://nodejs.org/api/modules.html#modules_loading_from_the_global_folders)) -- throws an error if it's not there.
3. This behavior is overridden with a stub during tests
4. When the blueprints are run (e.g. during `ember install ember-electron`), `ember-electron` uses the `electron-forge` API directly since (presumably) there isn't anything else yet. https://github.com/adopted-ember-addons/ember-electron/blob/8605e08965d159713f74bb2885b106a0d5d77d8d/blueprints/ember-electron/index.js#L2

lib/utils/forge-core.js
jacobq: Description > Used as a function:

> When Error is used like a function -- without new, it will return an Error object. Therefore, a mere call to Error will produce the same output that constructing an Error object via the new keyword would.

jacobq: I just peeked at Ember.js' source and see that they use/prefer `new Error` so I guess I'm in the minority. (I like to remove code that I don't think adds meaning so always omit the `new` for `Error`) Just wondered if you had some seasoned opinion that I could learn from.

bendemboski: 🤷‍♂ I guess I don't like taking advantage of unique behaviors when standard behaviors apply since I think it makes code harder to read by adding mental load (have to remember that `Error`, unlike other constructors, can be invoked without `new`). Also makes it harder for people coming from other languages.

jacobq: Fair enough -- thanks for taking a minute to share your thoughts.

PR #433, opened 2019-12-07
feat: check dependencies in Electron project

bendemboski:
We now use ember-cli-dependency-checker to check that dependencies in the Electron project are up-to-date before running any of our commands. This also involved a little bit of a refactor of the common pre-command functionality.


lib/utils/prepare-run-command.js
```diff
@@ -0,0 +1,28 @@
+const Win = require('ember-cli/lib/utilities/windows-admin');
+const fs = require('fs');
+const { electronProjectPath } = require('../utils/build-paths');
+const checkDependencies = require('../utils/check-dependencies');
+const { upgradingUrl } = require('../utils/documentation-urls');
+const SilentError = require('silent-error');
+
+module.exports = async function prepareRun(project) {
+  if (!fs.existsSync(electronProjectPath)) {
```
jacobq: Feels a little weird to me to be using the sync versions of `fs` methods in an async function, but I realize `exists` is a tricky one, and admittedly `existsSync` seems more readable than:

```js
const fsAccess = util.promisify(fs.access);
// ...
await fsAccess(filePath, fs.F_OK);
```

PR #434, opened 2019-12-07
fix: update .travis.yml correctly from the blueprint

### LINKED ISSUE #429: Failed to update .travis.yml on new/empty project
Attempting to run 3.0.0-beta.0 on a new/empty Ember app results in an error/warning like this:

```
Updating .travis.yml
Failed to update .travis.yml. For info on manually updating your CI config read https://adopted-ember-addons.github.io/ember-electron/docs/guides/ci.
Error: TypeError: Cannot read property 'includes' of undefined
```

I think this because the [default `.travis.yml`](https://github.com/ember-cli/ember-new-output/blob/master/.travis.yml) does not have an `install` section, so the following line reduces to `let entry = [].find(...) || [].find(...)` which is `undefined`.

https://github.com/adopted-ember-addons/ember-electron/blob/a6dd363edbf2d53f5b300502912ad04051ab5f48/blueprints/ember-electron/index.js#L87

bendemboski:
This code was buggy and not working, so fix it and verify it in the end-to-end tests

Fixes #429

### DIFF
--- a/blueprints/ember-electron/index.js
+++ b/blueprints/ember-electron/index.js
@@ -83,17 +83,20 @@ module.exports = class EmberElectronBlueprint extends Blueprint {
 
       // add install commands -- install dependencies in electron-app project,
       // and export display and launch xvfb
+      let hasInstallSection = Boolean(doc.install);
       doc.install = doc.install || [];
-      let entry = doc.install.find(entry => entry.includes('yarn ') || entry.includes('npm '));
-      if (entry.includes('yarn')) {
+      let usesYarn = Boolean(doc.install.find(entry => entry.includes('yarn ')));
+
+      if (usesYarn) {
         doc.install.push('__yarn_install__');
       } else {
+        if (!hasInstallSection) {
+          // If no install section is specified, travis defaults to running
+          // `npm install`, so if we're adding an install section, we need to
+          // explicitly add that since it will no longer be run by default.
+          doc.install.push('__npm_install_root__');
+        }
         doc.install.push('__npm_install__');
-      }
-
-      if (!doc.install.find(entry => entry.toLowerCase().includes('xvfb'))) {
-        doc.install.push('__export_display__');
-        doc.install.push('__xvfb__');
       }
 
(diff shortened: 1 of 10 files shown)


blueprints/ember-electron/index.js
```diff
@@ -96,18 +96,13 @@ module.exports = class EmberElectronBlueprint extends Blueprint {
...
-        doc.install.push('__xvfb__');
-      }
-
       // also, yawn quotes strings with certain characters in them even though
       // it isn't necessary, and it makes it harder to read. So we add
       // placeholders that won't be quoted and replace them in the output string
       yawn.json = doc;
       let output = yawn.yaml;
-      output = output.replace('__yarn__install__', `cd electron-app && yarn`);
-      output = output.replace('__npm__install__', `cd electron-app && npm install`);
```
jacobq: I had to stare at this diff for a minute to figure out what changed. Nice catch with the extra `_`

blueprints/ember-electron/index.js
```diff
@@ -84,8 +84,8 @@ module.exports = class EmberElectronBlueprint extends Blueprint {
       // add install commands -- install dependencies in electron-app project,
       // and export display and launch xvfb
       doc.install = doc.install || [];
-      let entry = doc.install.find(entry => entry.includes('yarn ') || entry.includes('npm '));
-      if (entry.includes('yarn')) {
+      let entry = doc.install.find(entry => entry.includes('yarn '));
+      if (entry) {
```
jacobq: Am I correct in understanding this to mean that we assume use of `npm` if we don't see `yarn`? (I think that's fine though could get pedantic and talk about [`bolt`](https://github.com/boltpkg/bolt) or something)

jacobq: Looks like this was fixed in the latest commits (with `hasInstallSection`)

jacobq:
> Doesn't this break projects with empty `install` sections...

Looks like this was fixed in the latest commits (i.e. with `hasInstallSection` logic)

jacobq:
Time for a fresh beta release?

PR #436, opened 2019-12-18
Update docs URL

bendemboski:
We've re-acquired ember-electron.js.org

### DIFF
--- a/lib/utils/documentation-urls.js
+++ b/lib/utils/documentation-urls.js
@@ -1,4 +1,6 @@
-const baseUrl = 'https://adopted-ember-addons.github.io/ember-electron/docs/';
+const { homepage } = require('../../package.json');
+
+const baseUrl = `${homepage}/docs/`;
 const guidesUrl = `${baseUrl}guides/`;
 const upgradingUrl = `${guidesUrl}upgrading`;
 const ciUrl = `${guidesUrl}ci`;
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ An Ember addon to turn Ember apps into cross-platform desktop applications, taki
 Documentation
 ------------------------------------------------------------------------------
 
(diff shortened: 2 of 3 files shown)


lib/utils/documentation-urls.js
```diff
@@ -1,4 +1,6 @@
-const baseUrl = 'https://adopted-ember-addons.github.io/ember-electron/docs/';
+const { homepage } = require('../../package.json');
```
rwwagner90: How does this work? The homepage should be a field in the package.json, not the entire package.json, right? I'm also not sure what this documentation urls util does? Do we need it since we have addon docs?

jacobq: The `{ ... }` does [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment)

rwwagner90: Ohhhh, wow you are right. It is early 😂. I still wonder if we need these doc utils at all? I would think not with addon docs?

jacobq: Maybe, but I think some are used in error/warning messages, which I would say has value.

rwwagner90: Ah okay

rwwagner90:
@bendemboski is ember-electron.js.org supposed to work? It appears blank.

rwwagner90:
@bendemboski @jacobq it is currently impossible to access any docs because https://ember-electron.js.org/ is blank and trying to go directly to ghpages just redirects there. This is why we removed the custom domain, as it would not work.

bendemboski:
Yeah, I had to add the CNAME file to get the js.org folks to approve the PR
to point the domain at us, which broke the old URL until we merge/deploy
this change.

PR #439, opened 2019-12-31
Fix Docs URL

devotox:


README.md
```diff
@@ -12,7 +12,7 @@ An Ember addon to turn Ember apps into cross-platform desktop applications, taki
 Documentation
 ------------------------------------------------------------------------------
 
-[View Docs](https://ember-electron.js.org/ember-electron/)
+[View Docs](https://ember-electron.js.org/docs/)
```
rwwagner90:
```suggestion
[View Docs](https://ember-electron.js.org/)
```

This should probably just be the root url, so they can land on the landing page, then click docs.

PR #444, opened 2020-04-27
Bump various deps, drop node 8 support

rwwagner90:
This bumps a lot of deps, many of which no longer support node 8, so I also changed our node engines to 10.x+.


rwwagner90:
@bendemboski @jacobq any ideas why the build is failing? I'm not familiar with the `electron make` tests.

bendemboski:
It basically just creates a new Ember project, installs `ember-electron` and then runs through the basic commands. So in theory you should be able to reproduce this by

```sh
$ ember new ee-test
$ cd ee-test
$ ember install ember-electron@git+ssh://... # git URL pointing to the failing build
$ ember electron:make
```

rwwagner90:
@bendemboski it looks like the problem is `Error: Could not find any make targets configured for the "linux" platform`. `yarn test` locally runs fine for me.

bendemboski:
Are you running locally on linux?

rwwagner90:
@bendemboski no, I am on MacOS.

rwwagner90:
Closing this and going to try updating deps in smaller chunks tomorrow.

PR #445, opened 2020-04-27
Ember 3.17/Octane updates

rwwagner90:


bendemboski:
Looks like the build failure is related to changes in `ember-cli` and how we mock the broccoli builder to test the `ember electron` command. It's blowing up [here](https://github.com/adopted-ember-addons/ember-electron/blob/master/node-tests/integration/commands/electron-test.js#L60) where we stub the internal method of the `ember-cli` watcher that creates the `broccoli-sane-watcher`, and make it return our mocked/instrumented watcher.

Based on [these changes](https://github.com/ember-cli/ember-cli/commit/097ab41516eb112a09f7ebaaeeab18bf34884011#diff-9ff56008b40607a31bc04df7f41906cbL29-L51) it looks like we need to change our test code to stub the `constructBroccoliWatcher` method instead of the `constructWatcher` method. Also, it looks like watcher interface is [slightly different](https://github.com/ember-cli/ember-cli/commit/097ab41516eb112a09f7ebaaeeab18bf34884011#diff-9ff56008b40607a31bc04df7f41906cbL169-R152) between the old watcher and new watcher, so we probably have to update our `MockWatcher` in that same test file to be something like

```javascript
class MockWatcher extends EventEmitter {
  constructor() {
    super(...arguments);
    this.currentBuild = {
      then: sinon.stub().callsFake((resolve) => resolve())
    };
  }
}
```

and hopefully that will get the test passing again?

rwwagner90:
Thanks for the tip @bendemboski, trying it now!

PR #447, opened 2020-04-28
Bump some deps

rwwagner90:


rwwagner90:
@bendemboski the version of inspector we had was fundamentally broken and so updating should be helpful. We have to have 3.13.x to include your PR with the fix https://github.com/emberjs/ember-inspector/pull/1084

Issue #425, opened 2019-11-26
Better Ember Inspector integration


bendemboski:
The API for installing extensions is a main process API, but we're doing it from the renderer processes via the Electron's `remote` module. This means that apps with `nodeIntegration` disabled will not be able to install the Ember Inspector, which is 100% as useful in non-`nodeIntegration` apps.

So I think we should do something like `electron-debug` does, and provide code that can be required from the main process to install the Electron Inspector, and add that code to `index.js` in the `ember-electron` blueprint/force template, so it would be something like:

```javascript
const { installEmberInspector } = require('ember-electron/extensions');
installEmberInspector();
```

Then it would be installed even with `nodeIntegration` disabled, and would allow users to opt out, and also learn about how install any other extensions they might want.

Open question is how to handle `devtron`. We could install it in the same manner, or point people to [electron-debug](https://github.com/sindresorhus/electron-debug), which I think is the most common way people install devtron.

rwwagner90:
I had some issues with electron-debug. It would be nice to not require electron-debug to use it, but it's also not a huge deal either way.

bendemboski:
@rwwagner90 whether `electron-debug` or not, my point is that devtron has nothing to do with Ember, and the problem of how to include it in an app is a problem that everybody with an Electron app has to solve, so I'd prefer to do it how everybody else does, whatever we decide that is (which is possibly doing nothing, and just providing info in an FAQ or something)

rwwagner90:
Sure, totally get that point. I suppose I don't see the downside in keeping it included though?

bendemboski:
Yeah, no big downside I don't think. Just maintenance/support cost of an increased "API surface area."

rwwagner90:
I would vote we keep it, as I like to have more magic, like Ember provides out of the box. The more we can bolt on, to allow people to not really need to understand Electron and still hit the ground running, the better. That's just my opinion though, and I don't feel super strongly about it, so removing is fine too.

bendemboski:
Yeah...I guess using the blueprint is a good compromise. It supplies the magic, but still allows users to opt out/customize. Let's leave it as an open question, and when this actually turns into a PR we can see how it comes together and discuss further.

jacobq:
BTW, how does one control the version of `ember-inspector` that's used? Seems like it's used by `ember-electron` but then also added to the app, which makes me think we have a similar situation as we did for electron-forge in 
[...]

PR #458, opened 2020-05-03
Install inspector via electron-devtools-installer

rwwagner90:
This is an attempt to use electron-devtools-installer to download the latest version of inspector, rather than worrying about keeping it up to date in our dependencies.

I was unable to yarn link to test this out, but would love some help figuring out how to test it!


bendemboski:
I suspect this isn't working because it's trying to use `electron-devtools-installer` from the renderer processes instead of the main process, and it doesn't look like `electron-devtools-installer` expects to do that.

I love the idea of using `electron-devtools-installer` and I think this is a great opportunity to also knock off #425 -- IIRC the main unknown/challenge there was figuring out a clean API that we would provide to easily install `ember-inspector`...but it looks like `electron-devtools-installer` is that API.

So I'd propose:
1. Add `electron-devtools-installer` to our forge template's [list of dependencies](https://github.com/adopted-ember-addons/ember-electron/blob/master/forge/template.js#L68) to install in the Electron app when generating it using the blueprint.
2. Add code to use `electron-devtools-installer` to install `ember-inspector` into the [electron-app/src/index.js](https://github.com/adopted-ember-addons/ember-electron/blob/master/forge/files/src/index.js) main process code that our forge template generates
3. Maybe add the same code to `lib/test-support/test-index.js` so `ember-inspector` is installed by default when running tests?

That way we'll be doing it from the main process, will solve the issue described in #425, will get `ember-electron` itself out of the business of managing this (just bootstrap the user's app so the user can manage it their self), will stop making the hard-to-work-around assumption that the user only wants `ember-inspector` installed when running in development mode, and will allow some nice cleanup:

1. Remove `ember-inspector` from `ember-electron`'s `dependencies` and from our forge template's list of `devDependencies` to install in the Electron app
2. Remove the `ember-inspector` installation code from that `vendor/install-extensions.js` file (so we only need a story for `devtron` to be able to get rid of it entirely

Note that if we do these cleanup steps, it will be a breaking change, but I think that's fine since we're still beta, as long as we provide some upgrade instructions, which I think should be pretty straightforward (`cd electron-app && yarn add electron-devtools-installer` and then paste this code snippet into your `src/index.js`)

rwwagner90:
@bendemboski thank you for the very detailed explanation! I think I did all the things suggested. I believe there is something about my app that causes the whole app to crash every time I open inspector, so I don't think I am the right person to test it out fully. It does seem to be fairly identical to what currently happens though.

rwwagner90:
Should we use something like https://github.com/sindresorhus/electron-is-dev to check if it is dev before installing inspector?

bendemboski:
> Should we use something like https://github.com/sindresorhus/electron-is-dev to check if it is dev before installing inspector?

Yeah, I suppose. Another blueprint/template dependency, but I think it's better than including it in production by default, and it is entirely in the user's control to make unconditional or use a different method of determining whether to enable it...

rwwagner90:
@bendemboski finally got the tests to pass! This should be ready for review now I believe.

forge/files/src/index.js
```diff
@@ -42,6 +44,12 @@ app.on('window-all-closed', () => {
 });
 
 app.on('ready', () => {
+  if (isDev) {
+    installExtension(EMBER_INSPECTOR)
+      .then((name) => console.log(`Added Extension:  ${name}`))
```
bendemboski, line 49: I don't think we really need to be logging this on every startup...just catch/console.error should be sufficient for logging, right?

lib/test-support/test-index.js
```diff
@@ -49,6 +50,10 @@ app.on('window-all-closed', function onWindowAllClosed() {
 });
 
 app.on('ready', function onReady() {
+  installExtension(EMBER_INSPECTOR)
+    .then((name) => console.log(`Added Extension:  ${name}`))
```
bendemboski, line 54: Same comment here

package.json
```diff
@@ -56,9 +56,10 @@
     "broccoli-string-replace": "^0.1.2",
     "chalk": "^4.0.0",
     "debug": "^4.1.1",
+    "electron-devtools-installer": "^3.0.0",
```
bendemboski, line 59: I wouldn't think we'd need these in `ember-electron`'s dependencies, because that would install it in the host Ember app, but we're already installing it in the Electron app using the blueprint.

vendor/install-extensions.js
```diff
@@ -2,8 +2,12 @@ function installExtensions() {
   try {
     window.requireNode('fs');
   } catch (e) {
-    console.warn('ember-electron is unable to require node modules, possibly because nodeIntegration is not enabled.');
-    console.warn('This prevents the installation of the Ember Inspector and Devtron extensions. Error:');
+    console.warn(
+      'ember-electron is unable to require node modules, possibly because nodeIntegration is not enabled.'
+    );
+    console.warn(
+      'This prevents the installation of the Ember Inspector and Devtron extensions. Error:'
```
bendemboski, line 9: It no longer prevents the installation of Ember Inspector! 🎉

rwwagner90: It might still prevent it though right? I don't know what electron-devtools-installer does behind the scenes, but I wasn't sure if maybe this was still relevant?

package.json
rwwagner90: Without this here, the tests would go 💥. I agree with you, but wasn't sure what our alternatives were.

forge/files/src/index.js
rwwagner90: I had just copied the logs from https://github.com/MarshallOfSound/electron-devtools-installer#usage but I'm fine with removing them if you think we should.

bendemboski: Yeah, it looks like it's just put there to make the usage more clear in the sample code -- I don't think we provide any value by telling the user each time they launch that we successfully installed the extension, and it will certainly annoy some users (at least one so far 😉)

vendor/install-extensions.js
bendemboski: No, I don't think so. The only reason that's a problem for this code is that here we use the remote module from the renderer process to install it, so if node integration is disabled, the remote module won't work.

But now we're doing it from the main process, so node integration has nothing to do with it, and if something causes an error installing it, that will show up in the error log in the catch block of `installExtensions(EMBER_INSPECTOR)`.

So let's change this to just talk about Devtron, since that's all we're installing here via the remote module after this change.

package.json
rwwagner90: Any ideas on how to make these not need to be dependencies, but still have the tests work?

bendemboski: Not offhand...how do the tests fail without it?

rwwagner90: Cannot find module electron-devtools-installer, etc. The dependencies have to be installed for testing, it seems.

bendemboski: Oh...right...because the blueprint generates `electron-app/tests/index.js` that does `require('ember-electron/lib/test-support/test-index')` and then the code running there will have a module search that is inside `ember-electron`, i.e. in the Ember app, not the Electron app.

I guess that's also why we need to have `electron-protocol-serve` as a dependency of `ember-electron`...ugh. That sucks because it means that the tests potentially run against a different version of things than the app (maybe not a huge deal for the devtools installer, but definitely not good for `electron-protocol-serve` and any other dependencies we might add in the future).

I guess the options I see are:
1. Move the code from that `test-index.js` module that lives in `ember-electron` into the blueprint-generated `tests/index.js` file.
2. Use `require.resolve()` to find the version of those dependencies that is in the electron app, using our knowledge of where it is that we get from `process.argv`
3. Leave it as-is and file an issue to address separately

I think I'd opt for a mix of (1) and (3), which would mean deleting the `installExtensions()` from `test-index.js` and adding it to `forge/files/tests/index.js` before the `require()`. Then at least we're not making the problem worse, and we can file an issue to sort out `electron-protocol-serve`, which will either result in doing something similar, or, I think more likely, getting rid of it entirely so we can have source maps post Electron 6.

rwwagner90: @bendemboski would it be okay with you if we handled this in a separate PR or did you want to do it in this one?

bendemboski: I think/hope it would just be a quick cut-and-paste job to move the code into the `forge/files/tests/index.js` and then remove these dependencies -- can we try that, and if it doesn't work, then revert and file an issue?

In either case, I really don't think we need `electron-is-dev` here since we're not running that check in the test version of the extensions installation, so I think either way we should pull that from the `ember-electron` deps.

rwwagner90: @bendemboski I don't understand what to copy over. Don't we need the app so we can do `app.on('ready'`? Does that mean I should require electron in the other file? I'm happy to do whatever you think is best, but I do not know what to do exactly.

bendemboski: Oh, okay, gotcha! Yeah, the test setup is a bit of a Rube Goldberg machine right now!

I'm saying to remove the code you added to `lib/test-support/test-index.js` and then make `forge/files/tests/index.js` look like

```javascript
const { app } = require('electron');
const { default: installExtension, EMBER_INSPECTOR } = require('electron-devtools-installer');

app.on('ready', function onReady() {
  installExtension(EMBER_INSPECTOR)
    .catch((err) => console.log('An error occurred: ', err));
});

require('ember-electron/lib/test-support/test-index');
```

I'm pretty sure that will allow us to remove the extra deps from `ember-electron`'s `package.json`.

### Explanation

When we run `ember electron:test` we build out the Ember app and instead of launching Electron and pointing to `/electron-app/src/index.js` as the main process code, we point to `/electron-app/tests/index.js` (both of these files are generated when the blueprint runs).

[forge/files/tests/index.js](https://github.com/adopted-ember-addons/ember-electron/blob/master/forge/files/tests/index.js) (which the blueprint uses to generate `/electron-app/tests/index.js`) is just one line:

```javascript
require('ember-electron/lib/test-support/test-index');
```

so it just executes exactly the code that lives in `/node_modules/ember-electron/lib/test-support/test-index.js`. But when that file tries to require modules, it's looking in `/node_modules` rather than `electron-app/node_modules`, so the dependencies need to be installed in the root Ember app rather than in the Electron app (`electron-app`).

So by moving the code to the blueprint-generated `electron-app/tests/index.js`, we allow it to require modules from `electron-app/node_modules`, i.e. the same set of modules that the non-test main process code has access to.

rwwagner90: Thanks for the explanation! I think I updated everything accordingly.

PR #465, opened 2020-05-07
Install devtron from the main process

rwwagner90:


rwwagner90:
@bendemboski docs updated!

PR #476, opened 2020-05-18
Switch from electron-protocol-serve to file: URLs

### LINKED ISSUE #442: Source maps not loading with Electron 7.x
When running `ember electron` with Electron 7.x, if you open the dev tools, the source maps do not load.

The root cause is changes in Chromium around the network service. Things broke in Electron 7.0.0, and [this](https://github.com/electron/electron/pull/21494) fix (a backport of [this](https://chromium-review.googlesource.com/c/chromium/src/+/1525270) chromium fix) got source maps working again for `file:` URLs. But they still don't work for custom protocols, like our `serve:` protocol that we use with `electron-protocol-serve`.

[Here](https://github.com/electron/electron/issues/21699) is the Electron issue I filed tracking this, but I have no idea if it will get attention/a fix anytime soon. Possible alternatives:

1. Look into a way of instrumenting the build or runtime protocol handling to rewrite the `sourceMappingURL` directive at the end of js files to use absolute `file:` URLs to point to the source maps (instead of relative URLs that end up resolving to custom protocol `serve:` URLs)
2. Toss the whole custom procotol/`electron-protocol-serve` thing, and go back to using `file:` URLs (which probably requires using the hash location type in Ember)

bendemboski:
`electron-protocol-serve` seems to have taken us too far off the beaten path, so this changes `ember-electron` to set up projects to load from file: URLs instead of the custom protocol. This requires a couple of workarounds to make Ember apps work properly when loaded from file: URLs -- all explained in detail in the new FAQ page. This gets us sourcemaps back in Electron 7+, and also removes the need for a variety of minor workarounds for security/sandboxing issues caused by our use of a custom protocol.

This PR also includes some related changes that simplify the infrastructure supporting `ember electron:test`, and improve our testing story a bit (i.e. the `ember-electron`'s own tests).

Fixes #442


bendemboski:
@rwwagner90 and @jacobq if you are able to try out these changes in your apps I'd very much appreciate it. I've tried them in mine, and they are working, but I'd like some more verification. Also, I'd love a verification that the [update instructions](https://github.com/adopted-ember-addons/ember-electron/blob/file-urls/tests/dummy/app/templates/docs/guides/upgrading.md#upgrading-from-300-beta2) from the previous beta work.

rwwagner90:
@bendemboski thanks for working on this! I will try it out in the next day or two.

tests/dummy/app/templates/docs/faq/routing-and-asset-loading.md
```diff
@@ -0,0 +1,101 @@
...
+
+So we have to go back to an older method of specifying logical locations, which is to use the URL hash. Fortunately, Ember supports this via the [HashLocation](https://api.emberjs.com/ember/release/classes/HashLocation) since that was the only option in older browsers. Since URL hashes don't affect how `file:` URLs are resolved to paths, we can tell the browser to load `file:///Users/me/projects/my-app/electron-app/ember-dist/index.html#my-route`, and it will load the `index.html` and then Ember can resolve the route using the path contained in the URL hash.
```
rwwagner90, line 19:
```suggestion
So we have to go back to an older method of specifying logical locations, which is to use the URL hash. Fortunately, Ember supports this via the [HashLocation](https://api.emberjs.com/ember/release/classes/HashLocation) since that was the only option in older browsers. Since URL hashes don't affect how `file:` URLs are resolved to paths, we can tell the browser to load `file:///Users/me/projects/my-app/electron-app/ember-dist/index.html#/my-route`, and it will load the `index.html` and then Ember can resolve the route using the path contained in the URL hash.
```

Hash routes need to be `index.html#/route-name`

forge/files/src/index.js
```diff
@@ -43,13 +26,22 @@ app.on('window-all-closed', () => {
   }
 });
 
-app.on('ready', () => {
+app.on('ready', async () => {
   if (isDev) {
-    require('devtron').install();
-    installExtension(EMBER_INSPECTOR)
-      .catch((err) => console.log('An error occurred: ', err));
+    try {
```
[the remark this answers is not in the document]
bendemboski: I don't feel too strongly about this, but it seems kinda lousy to prevent the whole app from launching if one of these development tools fails to load for whatever reason. Also, if something does fail, who knows what the error will look like and if it will be easy for users to figure out what's going on...

So, that was my thinking -- control the messaging and make it non-fatal.

forge/files/src/index.js
```diff
@@ -1,32 +1,15 @@
...
 const { default: installExtension, EMBER_INSPECTOR } = require('electron-devtools-installer');
+const { pathToFileURL } = require('url');
+const { app, BrowserWindow } = require('electron');
+const path = require('path');
+const isDev = require('electron-is-dev');
+const handleFileUrls = require('./handle-file-urls');
 
-let mainWindow = null;
+const emberAppDir = path.resolve(__dirname, '..', 'ember-dist');
+const emberAppURL = pathToFileURL(path.join(emberAppDir, 'index.html')).toString();
```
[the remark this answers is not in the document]
bendemboski: Oh, just a stylistic thing I guess. So many of the methods in the `fs` module are pretty ambiguously named -- `join`, `resolve', `parse`, `normalize` -- and the name of the module is only two characters, so I generally don't destructure it because I find code more readable with the `fs.` prepended everywhere so it's immediately obvious what the method actually is.

forge/files/src/index.js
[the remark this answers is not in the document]
rwwagner90: Sure, we can keep it 👍

blueprints/ember-electron/index.js
```diff
@@ -45,12 +46,54 @@ module.exports = class EmberElectronBlueprint extends Blueprint {
...
+    } else {
+      this.ui.writeLine(chalk.yellow([
+        `\nUnable to update rootURL setting to`,
+        `\`process.env.EMBER_CLI_ELECTRON ? '' : <previous value>\`,`,
+        `which is needed for your Ember app to load assets under Electron.`,
+        `See ${routingAndAssetLoadingUrl} for more information.`
+      ].join(' ')));
+    }
+
+    const locationTypeRegex = /(\slocationType:)/;
```
jacobq, line 73: This is a bit pedantic, but whitespace is allowed between the key name and the `:` so it might be worth adding `\s?` just before to cover that case also.

```js
[" foo:123", " foo : 123" ].map(s => /(\sfoo:)/.test(s)) // --> [ true, false ]
[" foo:123", " foo : 123" ].map(s => /(\sfoo\s?:)/.test(s)) // --> [ true, true ]
```

forge/files/src/handle-file-urls.js
```diff
@@ -0,0 +1,29 @@
+const { protocol } = require('electron');
+const { fileURLToPath } = require('url');
+const path = require('path');
+const fs = require('fs');
+const { promisify } = require('util');
+
+const access = promisify(fs.access);
```
jacobq, line 7: Remind me again why we can't use [`fsPromises`](https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_fspromises_access_path_mode)... Need to support Node < 10.x?

forge/files/src/index.js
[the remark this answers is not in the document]
jacobq: I agree that path's method names tend to be a bit too vague by themselves. In the past I've done things like `const { join: pathJoin } = requireNode('path')` thinking it might help with optimization / tree shaking / whatever, but after a while I quit as it just seemed to distract me visually and make me think too much.

blueprints/ember-electron/index.js
bendemboski: Yeah, I suppose. I was mainly trying to target the output of the default app blueprints. The leading `\s` was to make sure that the `locationType` wasn't a substring, like if the user had another setting that was `colocationType` or something...probably even less likely than the user putting spaces before the `:` 😄 I don't want to go crazy trying to handle all kinds of cases of modifications to the app-blueprint-generated, but I think that extra `\s*` is super reasonable -- I'll add it.

forge/files/src/handle-file-urls.js
bendemboski: Uh, yeah, we kinda recently dropped node 8 support, so that's why we were avoiding it before. However, this is a different case, because this code runs in the Electron main process, not the user's Node environment, so the `engines` in `package.json` isn't relevant here, just the version of Electron that the user is running.

But Electron has been on Node 10+ since 3.x, and I'm quite confident that there are other reasons this code would break on Electron <=2, not the least of which is the usage of `fileURLToPath` in this same file, which wasn't introduced until Node 10.12/Electron 5 😆 

I don't think we have an official policy on which versions of Electron we do and don't target, but since:
1. This is all blueprint-generated code that the user could modify/replace if they need to
2. We're in the midst of a major version bump

I think we're fine keeping the usage of `fileURLToPath` and using `fs.promises`. I'll make the change.

forge/files/src/index.js
```diff
@@ -43,13 +26,22 @@ app.on('window-all-closed', () => {
...
-app.on('ready', () => {
+app.on('ready', async () => {
   if (isDev) {
-    require('devtron').install();
-    installExtension(EMBER_INSPECTOR)
-      .catch((err) => console.log('An error occurred: ', err));
+    try {
+      require('devtron').install();
+    } catch (err) {
+      console.log('Failed to install Devtrom: ', err);
```
jacobq, line 34: Typo? Devtro**m** -> Devtro**n**

P.S. Thank you for making this message more meaningful. I hadn't personally ran into it, but overly generic error messages drive me nuts in logs.

bendemboski: Haha, yeah, I noticed that myself and have it fixed locally (new commit coming soon)

forge/files/tests/index.js
```diff
@@ -1,9 +1,30 @@
...
-  require('devtron').install();
-  installExtension(EMBER_INSPECTOR)
-    .catch((err) => console.log('An error occurred: ', err));
+const emberAppDir = path.resolve(__dirname, '..', 'ember-test');
+
+app.on('ready', async function onReady() {
+  try {
+    require('devtron').install();
+  } catch (err) {
+    console.log('Failed to install Devtrom: ', err);
```
jacobq, line 13: Same typo as previously: Devtro**m** -> Devtro**n**

forge/files/tests/index.js
```diff
@@ -1,9 +1,30 @@
...
+  }
+  try {
+    await installExtension(EMBER_INSPECTOR);
+  } catch (err) {
+    console.log('Failed to install Ember Inspector: ', err)
```
jacobq, line 18: Same missing ; as before...

lib/test-runner.js
```diff
@@ -5,7 +5,7 @@
 // // testem.js
 // module.exports = {
 //   "launchers": {
-//     "Electron": require("ember-electron/lib/test-support/test-runner")
+//     "Electron": require("ember-electron/lib/test-runner")
```
jacobq, line 8: No objection but curious why `test-support/test-runner` was changed to `test-runner`. Just tidying up?

jacobq: Disregard -- just noticed notes in commit message
https://github.com/adopted-ember-addons/ember-electron/commit/e4a719b0e6e131fe4c02a110bd21840bc2932edf

forge/files/src/handle-file-urls.js
rwwagner90: I think we should maybe support Electron 5+ or whatever we had to update Electron Forge for?

blueprints/ember-electron/index.js
jacobq: Got to draw the line somewhere, but for me that point is more like supporting `{ [eval("'foo'")]: ...}` 😱

lib/test-runner.js
```diff
@@ -21,50 +21,27 @@
 
 module.exports = {
   'exe': 'node',
-  'args': [__filename, '<cwd>', '<baseUrl>', '<testPage>', '<id>'],
+  'args': [__filename, '<testPage>', '<baseUrl>', '<id>'],
```
jacobq, line 24: I don't actually know what this is / what it needs to agree with. I assume it either goes with a `require` in testem's launchers or electron-forge's `start` below. Do we define the order of these args or is that specified by something else (e.g. testem)?

jacobq: Oh, does this go with the following?

```js
// These are the command-line arguments passed to us by test-runner.js
const [ , , , testPageURL, testemURL, testemId ] = process.argv;
```

bendemboski: Yeah, I'll add a comment

node-tests/fixtures/ember-test/test-index-extra.js
```diff
@@ -1,18 +1,24 @@
...
+      if (!Object.prototype.hasOwnProperty.call(extensions, 'devtron') || !Object.prototype.hasOwnProperty.call(extensions, 'Ember Inspector')) {
+        console.error('devtron and/or ember-inspector not installed', Object.keys(extensions)); // eslint-disable-line no-console
+        app.exit(-1);
+      }
+    });
   });
-});
+}
+
+testIndexExtra();
```
jacobq, line 24: Would we ever export or reuse this function? I'm wondering if there's a reason (other than stylistic preference, which is a reason I'm OK with in your case :D) not to use an IIFE instead.

bendemboski: Just stylistic -- I've always found IIFEs a bit hard to read, so generally only use them if I need them to generate a value in an object literal or something...no strong preference, though.

node-tests/unit/blueprint-test.js
```diff
@@ -25,16 +26,53 @@ describe('blueprint', function() {
...
+
+    async function getEnv() {
+      let environmentJs = path.join(__dirname, '..', 'fixtures', 'config-environment', 'environment.js');
+
+      mkdirSync('config');
+      copyFileSync(environmentJs, path.join('config', 'environment.js'));
+
+      await blueprint.updateEnvironmentConfig();
+
+      let factory = require(path.join(process.cwd(), 'config', 'environment.js'));
```
jacobq, line 48: Another nit: can we reduce the `<cwd>/config/environment.js` duplication? e.g.

```js
mkdirSync('config');
const newEnvJsPath = path.join(process.cwd(), 'config', 'environment.js');
copyFileSync(environmentJs, newEnvJsPath);
await blueprint.updateEnvironmentConfig();
let factory = require(newEnvJsPath);
```

tests/dummy/app/templates/docs/faq/common-issues.md
```diff
@@ -1,37 +1,5 @@
...
-
-You should carefully consider if doing this is safe, as it potentially opens up new security holes that vanilla Electron using `file:` URLs does not. The difference is that the `allow_universal_access_from_file_urls` option evaluates the situation on each request, and if the current origin isn't a `file:` URL, it will enforce the same-origin policy. When using `webPreferences` you are disabling the same-origin policy for the entire window, so if the window somehow navigates to an `http:` URL, the same-origin policy will still be disabled. So be sure to consider your circumstances carefully from a security standpoint when doing this, and ensure that if your window needs to display remote or untrusted content, it does so in a sandboxed `WebView`.
-
-
 ## I can't open the Ember Inspector in the packaged app
 
 This is known and desired behavior of Electron. The Developer Tools are still accessible programmatically or via a shortcut (see below). However, the packages needed for [DevTron](https://github.com/electron/devtron) and the [Ember Inspector](https://github.com/emberjs/ember-inspector) addons are not currently bundled with production builds.
```
jacobq, line 5: Side question: weren't we talking about changing this behavior? Maybe I'm thinking of something else (can't find the issue I thought I remembered seeing....) Not that it would belong in this PR, but I can't remember if this info is still accurate.

bendemboski: Yeah, it is still accurate, although changing it is super easy, and we could provide instructions for making them always accessible (just move the packages from `devDependencies` to `dependencies` and then get rid of the `isDev` check around them).

If you think it's worthwhile, feel free to file an issue.

tests/dummy/app/templates/docs/faq/routing-and-asset-loading.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Source maps
+
+One might observe that the solution for images, fonts, etc., could also apply to the Javascript and CSS, and wonder why we need both solutions. The answer is source maps. Unfortunately when the Chromium developer tools load source maps, it is not intercept-able by the Electron `protocol` module, and the developer tools are not aware of any intercepting that happens in the main process. So if we left the leading `/` on the script tags, e.g. `<script src="/assets/vendor.js"></script>`, then when the developer tools sees the `sourceMappingURL=vendor.map` at the end of `vendor.js` it will treat `vendor.map` as a path that is relative to where it thinks the script was loaded from, i.e. `/assets/vendor.js`, so it will try to load it from `/assets/vendor.map`. Since this request isn't intercept-able, the source maps will fail to load. So to make source maps work, we really need to load the Javascript and CSS files from relative URLs, and since we cannot easily rewrite the other assets' URLs to be relative, we need two different solutions for the two different flavors of assets.
```
jacobq, line 55: 👍 This answered the question I had from the beginning of this FAQ article. Thank you for including these details.

tests/dummy/app/templates/docs/guides/upgrading.md
```diff
@@ -75,25 +75,7 @@ Either way, you will need to update the configuration since the format has chang
...
-while `ember-electron` 3.x's contains:
-
-```
-protocolServe({
-  cwd: join(__dirname || resolve(dirname('')), '..', 'ember-dist'),
-  app,
-  protocol,
-});
-```
+The equivalent of `ember-electron/main.js` is `electron-app/src/index.js`, so you'll want to put the contents of your `main.js` there (note that `src/index.js` is specified by the `main` entry in `electron-app/package.json`, so you can name it whatever/put it wherever you want as long as you update the `main` entry accordingy). Because the Electron project structure differs a little from `ember-electron` 2.x's, you'll need to replace anywhere you reference the path to the Ember application directory (`../ember`) with the new path (`../ember-dist`).
```
jacobq, line 78: Typo: accordingy -> according**l**y*

tests/dummy/app/templates/docs/guides/upgrading.md
```diff
@@ -125,3 +107,28 @@ If you need to exclude files for other platforms from your packaged build, you c
...
+
+Since 2.x, we have removed `electron-protocol-serve` from the default blueprint in favor of loading the Ember app using `file:` URLs. The instructions above should get you set up to run properly, but if you app has any references/assumptions around the URL used to load the Ember app, you'll need to update them. If you added `webSecurity: false` to work around issues caused by `electron-protocol-serve` (as described [here](/versions/v2.10.2/docs/faq/common-issues#what-is-electron-protocol-serve-and-why-do-i-need-this-)) you should be able to remove it now.
+
+You can read more about the removal of `electron-protocol-serve` and loading from `file:` URLs [here](../faq/routing-and-asset-loading).
+
+# Upgrading from 3.0.0-beta.2
+
+Between `3.0.0-beta.2` and `3.0.0-beta.3` we removed `electron-protocol-serve` from the default blueprint as explained [here](#electron-protocol-serve). The best way to upgrade from a `3.0.0` beta version before `3.0.0-beta.3` is to:
+
+1. Start with a clean working directory
```
jacobq, line 121: Is clean working *directory* the same as clean working *tree*?

bendemboski: I'll clarify

forge/files/src/handle-file-urls.js
jacobq: FWIW the [Electron team only supports the latest 3 stable branches](https://www.electronjs.org/docs/tutorial/support#supported-versions). So now that 9.0.0 is out I think that means versions older than 7.x are no longer supported. I am fine with 5+, but my gut would be to target actively supported versions.

bendemboski: *grumble grumble* https://travis-ci.com/github/adopted-ember-addons/ember-electron/jobs/337264506#L610

jacobq:
:D
Change engines to ^10.17.0 || 12.* || >= 14?

jacobq:
Curious about the [electron 9 breakage with Devtron](https://travis-ci.com/github/adopted-ember-addons/ember-electron/jobs/337292292#L712-L716). Looks like we're seeing `TypeError: electron.BrowserWindow.addDevToolsExtension is not a function`. When I tried to build my app (using ember-electron@3.0.0-beta.2) with electron 9, I saw a test failure in CI like the following, does this sound related?

> Loading extension at /my-app/electron-app/node_modules/devtron failed with: The 'manifest_version' key must be present and set to 2 (without quotes). See developer.chrome.com/extensions/manifestVersion.html for details.

bendemboski:
Haha, they made my life super difficult by releasing Electron 9 in the middle of this PR 😆 

There are two issues at play, [this one](https://github.com/electron/electron/issues/23656) which affects both Ember Inspector and devtron, and is failing the tests (but has an easy workaround) and [this one](https://github.com/electron/electron/issues/23676) which I just filed, which only affects devtron, and which I have no idea if there's a workaround...seems like not.

Looking in `devtron`, its manifest doesn't look at all like I think a manifest for an extension should...or it's a super duper duper old format that maybe Chromium and/or Electron dropped support for? Anyway, nothing to do with how we install it -- it's just broken on Electron 9.

bendemboski:
@rwwagner90 what can we do about the codeclimate failure? I don't agree that addressing any of the issues would actually make the code clearer, although happy to discuss -- how do we get the checks passing?

rwwagner90:
@bendemboski you should be able to log in and approve them. I logged in and approved them, so should be good to go!

forge/files/src/handle-file-urls.js
rwwagner90: I would be fine with Electron 7+, and I would honestly also be fine with bumping node minimum to 12, but ember-cli supports 10 until April 2021.

bendemboski: Yeah, this is kinda silly because `eslint` is warning us about code that runs in Electron, not the environment that the `engines` field is describing. So I don't think bumping our minimum required node version makes any sense because I don't think we use any `fs.promises` in code that actually runs in the user's node environment, so we'd be dropping support just to make the linter happy.

That being said, I do think it makes sense to get more specific about what Electron versions we support, but haven't put any thought into really how to accomplish that -- I guess it would just be documentation and then testing against all supported (major) versions?

bendemboski:
@rwwagner90 thanks!

forge/files/src/handle-file-urls.js
rwwagner90: Yeah, I think we would mainly list in the documentation that we don't support earlier Electron versions, but that it may still work.

PR #490, opened 2020-06-01
Correct the link to felixs github

spruce:


jacobq:
Thanks for catching this

PR #555, opened 2020-08-16
Restore require() sooner

bendemboski:
In complicated ember-auto-import/webpack scenarios where the desire is for webpacked modules to be able to require() node modules, setting up our require() wrapper in shim-footer.js might be too late, so set up the wrapper sooner -- in vendor.js. It will still be after Ember has loaded, so it won't interfere with Ember setting up its module loader.

### DIFF
--- a/index.js
+++ b/index.js
@@ -23,6 +23,14 @@ module.exports = {
     };
   },
 
+  included(app) {
+    this._super.included.apply(this, arguments);
+
+    app.import('vendor/wrap-require.js', {
+      type: 'vendor',
+    });
+  },
+
   contentFor(type) {
     const { env: { EMBER_CLI_ELECTRON } } = process;
 
--- a/lib/resources/shim-footer.js
+++ b/lib/resources/shim-footer.js
@@ -1,25 +1,6 @@
 // ember-electron
 ((win) => {
-  const { requireNode, require: requireAMD } = win;
-
-  // Redefine a global `require` function that can satisfy both Node and AMD module systems.
-  if (requireAMD) {
-    win.require = (...args) => {
-      try {
-        return requireAMD(...args);
-      } catch(error) {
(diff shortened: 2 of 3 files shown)


bendemboski:
I don't think it should really have any (although I'm not certain of course because of all the complexity here). This is what `index.html` (from the default blueprint) roughly looks like, with some comments describing the change:

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <title></title>
    <meta name="description" content="">
    <meta name="viewport" content="width=device-width, initial-scale=1">

    <!--
      here we did/do inject a script tag (shim-head.js) that saves off the NodeJS
      globals (module, process, and require) and then deletes them off of the window
      so they aren't there to confuse Ember while it sets up its module loader and
      whatnot
    -->
    {{content-for "head"}}

    <link integrity="" rel="stylesheet" href="{{rootURL}}assets/vendor.css">
    <link integrity="" rel="stylesheet" href="{{rootURL}}assets/groening.css">

    {{content-for "head-footer"}}
  </head>
  <body>
    {{content-for "body"}}

    <!--
      After this change, somewhere in vendor.js is where we restore the
      NodeJS globals. We don't know exactly where it will be, except it will
      always be after Ember sets up its module loader because of how the
      vendor script concatenation works, and it will now be before some of
      the vendor code, and before all of the application code in app-name.js,
      meaning that it won't interfere with setting up the module loader, but
      will be back to "normal" for as much code as possible, including all of
      the application's code
    -->
    <script src="{{rootURL}}assets/vendor.js"></script>
    <script src="{{rootURL}}assets/app-name.js"></script>

    <!--
      Here is where we used to restore the NodeJS globals
    -->
    {{content-for "body-footer"}}
  </body>
</html>
```

PR #671, opened 2020-12-14
Convert to GitHub actions

rwwagner90:
@bendemboski I am not sure if this does all the same things as Travis, but I think it should. Let me know what you think.


bendemboski (approved):
Looks good...I wonder how hard it would be to test on Windows also so we can move away from appveyor and have all the CI in one place.

rwwagner90:
@bendemboski moving Windows here sounds good to me. We should test on macOS, Linux, and Windows, but that can be a follow up.

PR #676, opened 2020-12-28
Set `contextIsolation` to false for tests

rwwagner90:
Not sure if we want this false or true, but it needs to be set to something.

### DIFF
--- a/lib/test-support/index.js
+++ b/lib/test-support/index.js
@@ -35,6 +35,7 @@ function openTestWindow(emberAppDir) {
     height: 600,
     webPreferences: {
       backgroundThrottling: false,
+      contextIsolation: false,
       nodeIntegration: true
     }
   });


bendemboski:
Why do we need to set this? I thought Electron took care of it for us -- disabled by default until Electron 12 and then enabled by default. Were you running into some kind of failure when not setting it either way?

rwwagner90:
@bendemboski yeah, there is a warning printed if you're running Electron 11+ I think and you set no value. They want you to take a stance either way. Since the default in 12 is to enable, perhaps we should make it true?

bendemboski:
This is a potentially breaking change though, no? As I understand it, it affects more than just preload scripts, so if anyone is using any of the features that context isolation disables, setting it to true would make their tests start to fail. Setting it to false is also problematic because then the tests are running with it disabled, while the actual app might have it enabled (or might switch to having it enabled when the user updates to Electron 12), so the tests might pass when the actual application would fail.

I think users would want to keep the setting in sync between their application windows and their testing windows, which I think means that this isn't a decision we can make for the user because we don't know the answer. That suggests that instead of supplying a value for `contextIsolation` on behalf of the user, we should make `openTestWindow()` accept it as an option, add a section to the documentation, and then leave it up to the user to handle the deprecation however they want. That way we're not trying to make any decisions for them.

I think maybe the best way to go about it would be to do the equivalent of (I can't remember if we can use rest/spread or not):

```javascript
function openTestWindow(emberAppDir, webPreferences) {
  let window = new BrowserWindow({
    width: 800,
    height: 600,
    webPreferences: {
      backgroundThrottling: false,
      nodeIntegration: true,
      ...webPreferences
    }
  });
  // ...
}
```

A sort-of counter-argument to the above is that we're already enabled `nodeIntegration` for them, so why not disable `contextIsolation` for them like your code does now despite the test/real app inconsistencies...I suppose that might not be too bad, I'm just more nervous about this case because unlike `nodeIntegration`, `contextIsolation` has differing behavior depending on the Electron version, so seems more likely to cause confusion/unexpected badness.

So happy to debate further, but this seems like a good opportunity to allow users to do their own handling of the `contextIsolation` change (meaning the blueprint-generated apps will by default print the warning until Electron 12 is released) and also fix the already-existing limitation of `openTestWindow()` where you're forced to use a fixed `webPreferences` configuration.

The other option would be to just do nothing and in the documentation tell users to either ignore the warning until Electron 12 ships, or replace their usage of `openTestWindow()` with a variant of the code it contains (which is only like 25 lines of code).

rwwagner90:
@bendemboski I'm open to whatever here. I think setting it to false could be okay, and we default to making the tests less safe basically, but more open. Then we could document how to change it or something. Not really sure. Perhaps we should stop shipping this for users and instead have blueprints copy it over?

bendemboski:
@rwwagner90 yeah, okay, I'm fine with that. Let's just set it to false (in line with how we set node integration to true) and we can think about providing documentation for how to change it, or tooling to make it easier to change, in the future if it ever seems like a priority.

bendemboski (approved):

PR #685, opened 2021-01-08
fix: replace deprecated 'crashed' event

knownasilya:
Basically 'crashed' is deprecated: https://www.electronjs.org/docs/api/web-contents#event-crashed-deprecated

### DIFF
--- a/forge/files/src/index.js
+++ b/forge/files/src/index.js
@@ -63,9 +63,7 @@ app.on('ready', async () => {
     if (details.reason === 'killed' || details.reason === 'clean-exit') {
       return;
     }
-    console.log('Your Ember app (or other code) in the main window has crashed.');
-    console.log('This is a serious issue that needs to be handled and/or debugged.');
-    console.log('Reason: ' + details.reason);
+    console.log('Your main window process has exited unexpectedly -- see https://www.electronjs.org/docs/api/web-contents#event-render-process-gone');
   });
 
   mainWindow.on('unresponsive', () => {
--- a/forge/files/src/index.js
+++ b/forge/files/src/index.js
@@ -59,9 +59,13 @@ app.on('ready', async () => {
     mainWindow.loadURL(emberAppURL);
   });
 
-  mainWindow.webContents.on('crashed', () => {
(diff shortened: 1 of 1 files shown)


rwwagner90:
@knownasilya do you happen to know what version of Electron it was deprecated in? We may want to take a stance on what Electron versions we support too.

knownasilya:
Well the new event was added here https://www.electronjs.org/releases/stable?version=8&page=2#release-notes-for-v840

knownasilya:
Also looks like deprecated in 8 https://github.com/electron/electron/pull/23096

forge/files/src/index.js
```diff
@@ -59,9 +59,13 @@ app.on('ready', async () => {
     mainWindow.loadURL(emberAppURL);
   });
 
-  mainWindow.webContents.on('crashed', () => {
+  mainWindow.webContents.on('render-process-gone', (_event, details) => {
+    if (details.reason === 'killed' || details.reason === 'clean-exit') {
+      return;
+    }
```
bendemboski, line 65: I think we should update the language here -- it kinda implies that it's a code problem, when it could be OOM or windows integrity or the user killing it intentionally with a SIGTERM, etc. Maybe we can just say `Your main window process has exited unexpectedly -- see <link to documentation for renderer-process-gone event>` and then also log out the reason.

bendemboski (changes requested):

bendemboski:
I'm not sure we need to take a version stance because of this. This code is in the blueprint, and the blueprint will always install the latest version of Electron. We've never said that the blueprint-generated code is compatible with past versions of Electron (that would incur a pretty big maintenance cost), so if users go in and customize their Electron version, it's completely reasonable for them to expect to have to update the blueprint-generated code to work with the older version as well.

I suppose we could mention this in the documentation somewhere, but I'm not even sure that's necessary...seems like an edge case that someone would install `ember-electron` without the intent to use the latest version of Electron.

knownasilya:
@bendemboski updated

bendemboski (approved):

PR #686, opened 2021-01-08
fix: following #676 to set contextIsolation explicitly

knownasilya:
See https://github.com/electron/electron/issues/23506

Decided to keep this false because of #676, but I think it should be true and now might be the time for that since it's still a beta.

### DIFF
--- a/forge/files/src/index.js
+++ b/forge/files/src/index.js
@@ -45,6 +45,10 @@ app.on('ready', async () => {
   mainWindow = new BrowserWindow({
     width: 800,
     height: 600,
+    webPreferences: {
+      // Details about what this is https://github.com/electron/electron/blob/master/docs/tutorial/context-isolation.md
+      contextIsolation: false,
+    },
   });
 
   // If you want to open up dev tools programmatically, call


rwwagner90:
@knownasilya I'm fine with making it true. I think we left it out intentionally, like we don't take a stance on `nodeIntegration`, so that people will get the console warning and have to pick a value for themselves. Making things more secure and requiring opting out seems like a good default though. @bendemboski thoughts?

bendemboski:
I'm not sure that we want to do this. I know we get a warning printed now, but it looks like Electron 12 is getting pretty close to being released, at which point it will apply a default without a warning. The purpose for the warning to to make sure that people don't run into problems updating the Electron version in their existing application, but this code is in the blueprint, so it only applies to new projects, not projects being updated.

Moreover, [electron-quick-start](https://github.com/electron/electron-quick-start) and [electron-forge's base template](https://github.com/electron-userland/electron-forge/blob/master/packages/template/base/tmpl/index.js) don't set it, so I don't think we should diverge from the practices of the rest of the Electron community.

So I think, like users of those two projects linked above, we should just live with the fact that our users will get deprecation warnings by default for the next few weeks or couple months or whatever until Electron 12 ships, rather than have a default baked in that duplicate's the framework's default or that we end up removing in the near future.

PR #699, opened 2021-01-22
Remove Mac from tests

rwwagner90:
Mac tests were having some issues, so removing them for now. We should definitely add back in the future, so we are testing on all platforms.


bendemboski:
@rwwagner90 did you try rerunning CI? It looks like it might just be some kind of transitory error? https://github.com/adopted-ember-addons/ember-electron/runs/1749036285#step:7:73

rwwagner90:
@bendemboski I wasn't sure, since we didn't test on Mac before. We can definitely add it back.

PR #704, opened 2021-01-25
Build on MacOS in CI

bendemboski:


bendemboski:
@rwwagner90 looks like the issue you were seeing was probably some transitory network thing 🎉

rwwagner90 (approved):

PR #747, opened 2021-03-02
pass allowFileAccess option to electron-devtools-installer

evoactivity:


forge/files/src/index.js
```diff
@@ -39,7 +39,9 @@ app.on('ready', async () => {
       console.log('Failed to install Devtron: ', err);
     }
     try {
-      await installExtension(EMBER_INSPECTOR);
+      await installExtension(EMBER_INSPECTOR, {
+        loadExtensionOptions: { allowFileAccess: true },
```
rwwagner90, line 43: Should this be `loadExtensionOptions: { allowFileAccess: true }` or just `{ allowFileAccess: true }`?

evoactivity: It's needed. `loadExtensionOptions` is what electron-devtools-installer passes to the options argument of electron's `loadExtension` method, which is where the allowFileAccess option is now available.

rwwagner90:
Thanks for the PR!

PR #751, opened 2021-03-03
Use process.execPath is the test runner

### LINKED ISSUE #750: Should use full path to `node` for launcher
https://github.com/adopted-ember-addons/ember-electron/blob/98e9237282ea572de3b490b26de0ebfc4c3eebad/lib/test-runner.js#L23

Should be:

```js
  exe: process.execPath,
```

bendemboski:
Use process.execPath instead of 'node' so we use the same Node executable that is running ember-cli, rather than rely on the user's PATH to point to the right (or any) version of Node.

Thanks @rwjblue for pointing this out

Fixes #750

### DIFF
--- a/lib/test-runner.js
+++ b/lib/test-runner.js
@@ -20,7 +20,7 @@
 //
 
 module.exports = {
-  exe: 'node',
+  exe: process.execPath,
   // These arguments are used in `lib/test-support/index.js`, which is called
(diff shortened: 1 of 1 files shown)


rwwagner90 (approved):

bendemboski:
Thanks @rwjblue!

Issue #763, opened 2021-03-14
requireNode is not a function even with nodeIntegration turned on


dtolios:
I just upgraded my app to ember-electron v3 and am having trouble getting "requireNode" working in my app. I have the following set in my index.js:

```
mainWindow = new BrowserWindow({
    width: 800,
    height: 600,
    webPreferences: {
      nodeIntegration: true,
    },
  });
```

I'm still having a problem the moment I launch my app because "requireNode is not a function". 

If i switch that to "require" I still get the same error, but coming from the ember-electron vendor script. 

Any tips or help is greatly appreciated as usual :)

rwwagner90:
@dtolios did you follow all the instructions here? https://ember-electron.js.org/docs/guides/upgrading

dtolios:
I did run that, but to verify I also spun up a new ember app doing the following:

1. Opened terminal in my projects/ dir
1. ran `ember new ember-electron-test`
1. ran `ember install ember-electron`
1. added the webPreferences with nodeIntegration to my main.js file
1. added a component which called requireNode("electron")
1. ran `ember electron`
1. saw error in my console

With that said, I've also been reading that I can wrap what I need using a preload script, but I'm also having a bit of trouble there, because the script doesn't seem to be run.

rwwagner90:
What file are you using `requireNode` in?

dtolios:
It was a component file located in:
```
app/
  components/
    my-button.js
```

dtolios:
Just a quick update, I was able to get around this in what I think is a nicer way this morning by using a preload script to wrap APIs. Here's what I did:

```
// in index.js of my electron app

// code...
mainWindow = new BrowserWindow({
    width: 1280,
    height: 720,
    frame: false,
    resizable: false,
    webPreferences: {
      preload: path.join(__dirname, "preload.js"),
    },
  });

// code...

  ipcMain.on("minimize", () => {
    mainWindow.minimize();
  });

  ipcMain.on("quit", () => {
    app.quit();
  });
```

```
// in preload.js
const { contextBridge, ipcRenderer, shell } = require("electron");

contextBridge.exposeInMainWorld("electron", {
  myOpenExternal: (url) => shell.openExternal(url),
  myMinimize: () => ipcRenderer.send("minimize"),
  myClose: () => ipcRenderer.send("quit"),
});
```

Then in my components i can just call things like:
```
window.electron.myMinimize();
```

[...]

PR #769, opened 2021-03-15
Include info on context isolation in documentation

bendemboski:
Our documentation on development and debugging that touches on node node integraton did not menton context isolation, which can show up as `requireNode()` appearing to just not work (see e.g. #763). So include a bit of info on context isolation and relevant links in this documentation section.


rwwagner90 (approved):
LGTM! Should we copy over the preload example as well or just let people read the docs and figure it out?

bendemboski:
I'm inclined not to because it feels more fully on the Electron side than the Ember/Electron integration side of things, but I could easily be swayed if you or anyone else feels differently.

Issue #637, opened 2020-11-06
ember electron:publish


basz:
I do not believe `ember electron:publish` is a supported command at the moment. 

To publish I need to do...

```
cd electron-app
NUCLEUS_TOKEN=xxxxxx npx electron-forge publish
```
Is this currently the recommended manner?

Is such a command planned as part of ember-electron?

rwwagner90:
I'm unfamiliar with `publish`. I use `ember electron:make` then upload my assets.

bendemboski:
Publish is not currently a supported command. After a brief look, I think the main thing that would need to be figured out would be supporting incremental builds, i.e. building and/or packaging and/or making, and then publishing based off of those already-built assets (like we do for the package and make commands).

The publish command has support for everything we need, so implementing support for publish would probably involve using the a copy of our make command and task as a starting point, figuring out the right `ember-cli` command options to expose in order to support the various scenarios of publishing different stages of pre-built assets (or building everything), and adapting the make command and task implementations accordingly.

PRs welcome!

basz:
Hmm. It seems you cannot run api.publish without it also doing the make process unless you provide makeResults to the api.publish. 

So would I be correct if we follow this logic.

We add a publish *task* that publishes with the makeResults so making is skipped.

we also add a flag to the make *task* to be able to run a publish *command* (ember electron:make --publish=true (defaults to false))

we can then run our make *command* as is but at the end do a publish where we provide the makeResult to the publish command.

(a bit later) I have tested passing and makeResults to api.publish his seems to work! 

`ember electron:publish` would then not exist but `ember electron:make --publish` would

bendemboski:
Fixed by #811

PR #811, opened 2021-04-29
feat: expose forge publish api

basz:
an attempt to expose the publish feature from electron forge

see #637

### DIFF
--- a/node-tests/integration/commands/make-test.js
+++ b/node-tests/integration/commands/make-test.js
@@ -19,12 +19,10 @@ describe('electron:make command', function () {
   mockElectronProject();
 
   let buildTaskStub;
-  let publishTaskStub;
   let command;
 
   beforeEach(function () {
     buildTaskStub = sinon.stub(BuildTask.prototype, 'run').resolves();
-    publishTaskStub = sinon.stub(PublishTask.prototype, 'run').resolves();
     sinon.stub(api, 'make').resolves();
     sinon.stub(api, 'publish').resolves();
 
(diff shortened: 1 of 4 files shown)


bendemboski (changes requested):
This looks great!

A couple of change requests:

1. We should expose the `publishTargets` option as a command argument, e.g. `ember electron:make --publish --publish-targets=@electron-forge/publisher-github,@electron-forge/publisher-s3` (or maybe `ember electron:make --publish --publish-target=@electron-forge/publisher-github --publish-target=@electron-forge/publisher-s3` would be cleaner? can't remember what `ember-cli`'s argument parsing makes easy and hard) because people might have different workflows for publishing the same project to different publisher targets.
2. Let's add some tests to the [make suite](https://github.com/adopted-ember-addons/ember-electron/blob/master/node-tests/integration/commands/make-test.js) to cover these new options. I think it should be pretty straightforward to stub forge's publish API like we already do for the make/package API and then verify that we do/don't call it correctly according to the arguments, and that we forward 0, 1, or >1 publish targets to the API when calling it., according to whichever of the `--publish-target` or `--publish-targets` command option we land on.

bendemboski:
Also, thinking through and writing down a few other things for posterity, but I don't think anything here is actionable:

### `electron:make` vs. `electron:publish`

We generally try to make the `ember-electron` command API mirror the `electron-forge` API as closely as possible, which means ideally we'd build an `electron:publish` command rather than adding a flag to the `electron:make` command. However, incremental workflow steps (e.g. reusing the output from previous steps) is also very important, and in this case `electron-forge` does not make that easy for us because the [MakeResults](https://js.electronforge.io/api/core/interfaces/publishoptions#makeresults) that we'd have to pass to the `publish` API to use the in-filesystem output of a previous `electron:make` command are non-trivial to generate from the filesystem, in particular determining the list of artifacts, which would vary from maker to maker.

Ideally `electron-forge` would do something like create a JSON file as part of the `make` output containing the `MakeResults` so the `publish` command could be pointed at a make output directory and read the data back in from the JSON file. But as long as that isn't the case, it doesn't seem feasible for us to create an `electron:publish` command that can be pointed at the output of a previous `electron:make` command, so adding a `--publish` flag to the make command seems like the best way to achieve that right now.

We could potentially also implement an `electron:publish` command that just can't skip the making step (put could reuse the output of a previous `electron:package` command, like `electron:make` can). Then it could expose all the dry run functionality that forge supports in the publish API (see below), and so when users need that they could use it (but would have to sacrifice the ability to reuse make output).

### Other publish options

I looked through the [options](https://js.electronforge.io/api/core/interfaces/publishoptions) that the `electron-forge` publish API accepts to see what we should support in the `electron:publish` command:

* `dir` -- set internally to point to `electron-app`
* `dryRun`/`dryRunResume` -- we could support these, but as mentioned above, it might be better to add them to an independent `electron:publish` command. Either way we can defer until someone actually wants it and we have known use case to target.
* `interactive` -- could make sense for an independent `publish` command, but not worth supporting now
* `makeOptions` -- not needed as long as `--publish` is a flag passed to `electron:make`
* `makeResults` -- see above, not feasible to pass as CLI options given how the forge API works now
* `outDir` -- only used for dry runs, so no value in supporting as long as we don't support the dry run arguments
* `publishTargets` -- see above

basz:
Q: How do I run the make suite tests locally?

bendemboski:
You should be able to run `yarn test-node-fast` to run all the unit/integration tests, or `yarn mocha <path to test file>` to run tests in a single file.

basz:
think I copied, stole, pasted, and adapted some tests... please let me know what you think

node-tests/integration/commands/make-test.js
```diff
@@ -157,4 +162,30 @@ describe('electron:make command', function () {
     rimraf.sync('electron-app');
     await expect(command.validateAndRun([])).to.be.rejected;
   });
+
+  it('can publish', async function () {
```
bendemboski, line 166: This is a great start on the tests. It's currently only testing the command code and not the task code, though, because the task is being stubbed rather than the forge API method. So let's stop stubbing the task and instead use the forge API stub to verify results. Then we can also verify that the `dir` argument is being passed correctly, and that the make results are being conveyed properly.

So, this test would look like:

```js
let makeResults = {};

// Change the make stub to return some fake results so we can ensure they
// they are passed to the publish API method
api.make.resetBehavior();
api.make.resolves(makeResults);

await expect(command.validateAndRun(['---publish'])).to.be.fulfilled;
expect(api.publish).to.be.calledOnce;
expect(api.publish.firstCall.args[0].dir).to.equal('electron-app');
expect(api.publish.firstCall.args[0].makeResults).to.equal(makeResults);
expect(api.publish.firstCall.args[0].publishTargets).to.be.undefined;
```

Then in the next two tests you can skip the make results stubbing since we've already verified that, and just assert that `api.publish` was called and that the `publishTargets` are correct.

bendemboski (changes requested):

basz:
of course, thank you for guiding me

bendemboski (approved):
Excellent, so glad to add publishing support -- thanks for the contribution!

Issue #914, opened 2021-08-24
Issue with shim loading on windows


basz:
Hi, I have an issue with the loading of the shim files on windows. ON MacOS they seem to load just fine.

Because of this requireNode (and thus ipcRenderer) is not available. At least I suspect they are missing becuase of this.

The path for the shims are on PC and Mac

`file:///D:/ember-electron/shim-head.js`

`file:///ember-electron/shim-head.js`

As you can see my app is launched from an additional drive called D.

If I inspect the interceptFileProtocol method I can see other url's (that do load fine) have a completely different pattern

for example:

`file:///D:/hf-app-lab-device-utility/electron-app/ember-dist/assets/vendor.js`

I I patch ember-electron/index.js injectScript function to not use a prepending slash as url

```
-  return `<script src="/ember-electron/${scriptName}"></script>`;
+  return `<script src="ember-electron/${scriptName}"></script>`;
```

the shims paths are correctly requested and interpreted (and my ipcRenderer works again).

`file:///D:/hf-app-lab-device-utility/electron-app/ember-dist/ember-electron/shim-head.js`.

evoactivity:
I just ran into this problem setting up a brand new project and had to make the same patch but on macos.

PR #915, opened 2021-08-24
fix: inject relative asset paths #914

basz:
windows seems to get confused when using absolute paths

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,5 @@
 function injectScript(scriptName) {
-  return `<script src="/ember-electron/${scriptName}"></script>`;
+  return `<script src="ember-electron/${scriptName}"></script>`;
 }
 
 module.exports = {


bendemboski:
Something more complicated is going on here, because if this were an across-the-board problem, then no `ember-electron` apps would ever be able to access `requireNode`, but we know of at least a few production `ember-electron` apps that work on windows.

Also, the same injection mechanism is used [here](https://github.com/adopted-ember-addons/ember-electron/blob/main/index.js#L41), and if that weren't load-able then the browser wouldn't be able to communicate with the testem server, and the end-to-end tests that run `ember electron:test` in a newly-generated app would fail in CI (that might be why this PR failed CI, although it failed on linux and didn't run the other scenarios).

I'm not sure what you're seeing and why, but I think we need a tighter diagnosis before making any code changes. Cross-platform file URI handling is a PITA, and `handle-file-urls.js` definitely has to jump through some hoops to make it work. So, I'd suggest that you:

1. Verify that you're up-to-date with the latest blueprint version of `handle-file-urls.js` (there was a fix for windows that went in some time ago, so this might be the whole issue right here)
2. If so, then do some logging to see exactly what's happening in `handle-file-urls.js`'s logic -- what each intermediate value is that leads to eventually generating the wrong filesystem path

rwwagner90:
I agree with @bendemboski that this likely needs more investigation. When I had windows issues before, it was because I had not updated the blueprints.

basz:
thanks for pointing out the blue print. after updating that the issue is resolved.

PR #1052, opened 2022-01-28
Run blueprint manually

bendemboski:
Unfortunately #1048 wasn't a full fix because `electron-forge` also has a dependency on `chalk`, so even without our own dependency, installing `ember-electron` could cause the hoisting to change triggering the underlying issue (https://github.com/ember-cli/ember-cli/issues/7431).

So, until that issue in `ember-cli` can be addressed, we're no longer running the blueprint automatically. Instead, we point the default blueprint at a new one that simply prints out instructions telling the user to run `ember g ember-electron` manually.

### DIFF
--- a/blueprints/ember-electron-postinstall/index.js
+++ b/blueprints/ember-electron-postinstall/index.js
@@ -1,0 +1,19 @@
+const Blueprint = require('ember-cli/lib/models/blueprint');
+
+module.exports = class EmberElectronBlueprint extends Blueprint {
+  constructor(options) {
+    super(options);
+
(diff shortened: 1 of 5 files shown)


bendemboski:
@rwwagner90 I think this is our solution for now. I created a new branch off of this and cherry-picked your ember-cli 4.1 commits, and [it passes](https://github.com/adopted-ember-addons/ember-electron/actions/runs/1763845745). So my proposal is we do this for now, i.e. instruct users to run the blueprint manually, until https://github.com/ember-cli/ember-cli/issues/7431 is addressed. Let me know what you think.

rwwagner90 (approved):

PR #1263, opened 2022-10-22
Ember 4.8

rwwagner90:


rwwagner90:
@bendemboski any idea why tests are failing here?

bendemboski:
@rwwagner90 I'd bet it's https://github.com/emberjs/ember-inspector/issues/2188. I had to disable installing Ember inspector in tests in my project because of that bug. The blueprint creates `test/index.js`, so I think we either need to wait for a fix for that issue to be released in Ember inspector, or figure out how to make the end-to-end tests comment out that code or something after running the blueprint but before launching the Ember tests.

rwwagner90:
@bendemboski gotcha, I will release the fix

PR #1279, opened 2022-11-04
electron-forge 6

bendemboski:


node-tests/acceptance/end-to-end-test.js
```diff
@@ -227,18 +227,19 @@ describe('end-to-end', function () {
...
-      // (e.g. rpmbuild). The default template specifies only darwin for the zip
-      // target, so we delete the targets setting so it's enabled for all targets
-      // the zip maker supports, which is all of them.
-      let packageJsonPath = path.join('electron-app', 'package.json');
-      let packageJson = readJsonSync(packageJsonPath);
-      let makers = packageJson.config.forge.makers;
-      let zipMaker = makers.find((m) => m.name === '@electron-forge/maker-zip');
-      delete zipMaker.platforms;
-      writeJsonSync(packageJsonPath, packageJson);
+    it.only('makes', () => {
```
rwwagner90, line 230: @bendemboski do we mean to be only running this one test?

bendemboski: Ack! Good catch!

rwwagner90 (approved):

bendemboski:
@rwwagner90 I actually don't think this is breaking...the only thing that looks to me like it has the potential to be breaking is the forge config...but I'm pretty sure it isn't. Previously you would either:

1. Have the config defined inline in your `package.json` under the `forge.config` key
2. Have the config in a `.js` file and in `package.json`, have the `config.forge` key contain a path to it

In either case, forge will still find the config. The only difference is that now with new forge, if you don't have a `forge.config` key, it will default to using `forge.config.js`. But previously if you didn't have a `forge.config` key in `package.json`, it just wouldn't work at all. So I think all upgrade paths here work.

rwwagner90:
@bendemboski Ah, the docs were confusing then. It looked like the changes you made indicated a config file was required instead of configuring in package.json.

bendemboski:
Yeah, it's confusing for sure 😆 But the documentation changes are all related to the blueprint/upgrading from 2.x instructions, which don't affect existing 3.x apps.

PR #1430, opened 2023-03-27
Add option to skip Ember build

deanylev:
Sometimes all I want to do is compile our native modules and run the app, without serving / rebuilding Ember (which takes forever)

### DIFF
--- a/lib/commands/electron.js
+++ b/lib/commands/electron.js
@@ -38,6 +38,12 @@ module.exports = BaseCommand.extend({
       aliases: ['lrprefix'],
       description: 'Default to _lr',
     },
+    {
+      name: 'skip-ember-build',
+      type: Boolean,
+      default: false,
+      description: 'Skip starting the Ember server and only start Electron'
+    },
   ],
 
   async run(commandOptions) {
--- a/lib/tasks/electron.js
+++ b/lib/tasks/electron.js
@@ -13,49 +13,57 @@ const { api } = require('../utils/forge-core');
 class ElectronTask extends Task {
   async run(options) {
     let { ui } = this;
+    let { environment, skipEmberBuild } = options;
 
-    ui.startProgress(chalk.green('Building'), chalk.green('.'));
+    if (skipEmberBuild) {
+      ui.writeLine(chalk.green('Skipping build...'))
+    } else {
+      ui.startProgress(chalk.green('Building'), chalk.green('.'));
+    }
 
-    let builder = new Builder({
+    let builder = skipEmberBuild ? null : new Builder({
       ui,
       outputPath: emberBuildPath,
-      environment: options.environment,
+      environment,
       project: this.project,
     });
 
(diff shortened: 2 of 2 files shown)


bendemboski:
Hi @deanylev thanks a lot for the PR! Did you know that you can do exactly this via `cd electron-app && yarn start`? Once the Ember app is built, `electron-app` is a fully prepared and ready-to-run `electron-forge` project that you can run without any Ember tooling.

This is why I've never taken the time to add a build-skipping option to the `ember electron` command. That being said, if you prefer to do it via the `ember electron` command, I don't have any objections. But, this task was written to mirror how `ember-cli`'s [serve task](https://github.com/ember-cli/ember-cli/blob/6c7a7b59f80d2a02f8d9844beb1c8f433ebbb558/lib/tasks/serve.js) is put together since they are close analogs of each other. The serve command accepts a `path` option, while the electron task just needs a `skipEmberBuild` option and has a hard-coded path, but aside from that I think it's all similar enough that you should be able to wire in the `skipEmberBuild` option using the same mechanisms as the serve task does (`MockWatcher`, `MockBuilder`, etc.) so we can keep the two looking very similar, and therefore keep it easier to maintain the `electron` task as the common APIs underlying it and the serve task change over time.

So if you're happy with the `cd electron-app && yarn start` solution, let's just go with that and close this PR, but if you still want the `skip-ember-build` option, let's make it happen! But I will request that you update how it's implemented to match the serve task.

bendemboski (changes requested):
Hi @deanylev thanks a lot for the PR! Did you know that you can do exactly this via `cd electron-app && yarn start`? Once the Ember app is built, `electron-app` is a fully prepared and ready-to-run `electron-forge` project that you can run without any Ember tooling.

This is why I've never taken the time to add a build-skipping option to the `ember electron` command. That being said, if you prefer to do it via the `ember electron` command, I don't have any objections. But, this task was written to mirror how `ember-cli`'s [serve task](https://github.com/ember-cli/ember-cli/blob/6c7a7b59f80d2a02f8d9844beb1c8f433ebbb558/lib/tasks/serve.js) is put together since they are close analogs of each other. The serve command accepts a `path` option, while the electron task just needs a `skipEmberBuild` option and has a hard-coded path, but aside from that I think it's all similar enough that you should be able to wire in the `skipEmberBuild` option using the same mechanisms as the serve task does (`MockWatcher`, `MockBuilder`, etc.) so we can keep the two looking very similar, and therefore keep it easier to maintain the `electron` task as the common APIs underlying it and the serve task change over time.

So if you're happy with the `cd electron-app && yarn start` solution, let's just go with that and close this PR, but if you still want the `skip-ember-build` option, let's make it happen! But I will request that you update how it's implemented to match the serve task.

deanylev:
> Did you know that you can do exactly this via `cd electron-app && yarn start`

For some reason I have always believed this doesn't work after `node-gyp clean` - you'd think I would've actually checked that before going to this trouble. Cheers!

PR #1508, opened 2023-06-15
Use absolute path

basz:
making via the Webpack Plugin requires an absolute path.

This PR will resolve any **dir** option to the forge api...

https://github.com/adopted-ember-addons/ember-electron/issues/1501


RobbieTheWagner:
4.2.2 released, thanks for the PR!

PR #1612, opened 2023-12-06
Switch to protocol.handle()

bendemboski:
In Electron 25, `protocol.interceptFileProtocol()` is [deprecated](https://www.electronjs.org/docs/latest/breaking-changes#deprecated-protocolregisterinterceptbufferstringstreamfilehttpprotocol), so update our template/blueprint file to use the new `protocol.handle()` if present, and fall back on `protocol.interceptFileProtocol()` if not.


RobbieTheWagner (approved):

RobbieTheWagner:
Thanks @bendemboski!

PR #1640, opened 2024-01-18
Remove electron-is-dev

bendemboski:
The package is no longer needed, and apparently switched to ESM, so was failing to load anyway

### DIFF
--- a/forge/files/src/index.js
+++ b/forge/files/src/index.js
@@ -6,7 +6,6 @@ const {
 const { pathToFileURL } = require('url');
 const { app, BrowserWindow } = require('electron');
 const path = require('path');
-const isDev = require('electron-is-dev');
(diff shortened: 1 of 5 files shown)


RobbieTheWagner (approved):

forge/files/src/index.js
```diff
@@ -32,7 +31,7 @@ app.on('window-all-closed', () => {
 });
 
 app.on('ready', async () => {
-  if (isDev) {
+  if (app.isPackaged) {
```
RobbieTheWagner, line 34: @bendemboski should this be `!app.isPackaged`? Seems like dev should be not packaged right? Maybe I am misunderstanding.

jagthedrummer: https://github.com/adopted-ember-addons/ember-electron/pull/1698

PR #1700, opened 2024-04-17
Switch to electron-extension-installer

bendemboski:
electron-devtools-installer is old and unmaintained and uses a no-longer-supported API.

Fixes https://github.com/adopted-ember-addons/ember-electron/issues/1697

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -93,7 +93,7 @@ module.exports = {
             allowModules: [
               'devtron',
               'electron',
-              'electron-devtools-installer',
+              'electron-extension-installer',
               'ember-electron',
             ],
           },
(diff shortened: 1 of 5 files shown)


RobbieTheWagner (approved):

RobbieTheWagner:
@bendemboski do we not need to install this package somewhere?

bendemboski:
@RobbieTheWagner these are all changes to the forge template which is run from the blueprint, so no impact on existing apps. We tell the forge template to install it in app [here](https://github.com/adopted-ember-addons/ember-electron/pull/1700/files#diff-388c178e7aeaa7596f7793bf460bfc1901ded315ef7df820a91774e263a8c610R68).

